PR #4, opened 2015-06-24
fixed projectGeneratorLegacy to work on OSX so all examples can be bu…

rezaali:
Hi! I cloned the most recent repo of openFrameworks and tried to generate all the examples via the projectGeneratorLegacy. It threw errors complaining about GLUT, I fixed the issues and was able to generate xcode projects for all the examples for OSX. 

This seems like the only way to generate the example for OSX if you are using the git version of openFrameworks, please correct me if I'm wrong! 

Thanks for OF 

<3
Reza


arturoc:
thanks!

PR #8, opened 2015-08-06
added changes for command line tool to allow for easier creation of p…

ofZach:
…rojects in an older style

@arturoc  this adds an env variable (PG_OF_PATH -- called this to not confuse with OF_PATH which is used in xcode xconfig files) as well as command line parsing of non named variables (ie, arguments that don't say -p="...").  This allows for you to create projects in the old style.


ofZach:
allows for 

screen shot 2015-08-06 at 8 22 20 am

to make 

screen shot 2015-08-06 at 8 22 27 am

if you set an enviroment variable up right.  assumes project creation now (is used to make you explicitly say "create" or "update") but doesn't do anything if no path or name (name = 1st non option argument) set.

not sure if the feedback to user is good enough, it would also be good to find a way to silence the warnings on path stuff like: 

    [ error ] ofFile: remove(): file does not exist
    [ error ] ofFile: copyTo(): source file does not exist

which make generally following the output of the command line tool hard....   I'll dig into this.

arturoc:
nice! yeah we could make the output a little bit more understandable and make the logging less verbose.

arturoc:
can you merge it with master? it's not mergeable any more, i did some changes to fix some problems with the nighlty build and i think that's conflicting now with your changes

ofZach:
yes will do hold on...

ofZach:
ok should be good now.  first rebase in my life, not sure what just happened.

arturoc:
:)

arturoc:
there's something not working when updating, even in master. if i create a project it contains the project file but when updating it will only copy the makefiles and some files depending on the platform but in linux the codeblocks files are missing and in osx the .xcodeproj is missing too. i guess it's trying to look for an old file and update it but if there's none, like it's the case when creating the packages it should copy it from the templates

ofZach:
sorry let me take a look...

ofZach:
are you missing the recursive flag?  I just tested the command line tool and it's working as expected.  It doesn't do recursive updating by default (for safety reasons) and if it doesn't see a src folder, you have to use force.  

commandLineTool -r pathToExamples

should work (I just tested on a folder with a few folders that have src folder inside them)

arturoc:
mmh, no what i mean is for example if i enter the allAddonsExamples dir and i do:

```bash
projectGenerator -o ../../.. .
```

it won't create the project files

arturoc:
or from examples/addons:

```bash
projectGenerator -o ../../.. allAddonsExample
```

ofZach:
ok let me check.  have so much respect for people who make command line apps now :)

arturoc:
actually that second option works, it's only when running it from the same folder where the project is when it won't create all the files

arturoc:
:)

ofZach:
ok I see the issue, give me a sec to fix it.  I seems like the paths are getting all mangled with the dot...   I feel like in the simplest case, we make need to handle this case differently ( and not join name to current path if it's a dot... ) but let me check on other relative paths, there might be something more sinister happening.

arturoc:
i'm quickly checking the code and projectPath is not used at all, it's always empty until that join so `projectPath = projectName` should do the same no?

arturoc:
also i think:

```

	void handleOFPath(const std::string& name, const std::string& value){
		
        ofLogNotice() << "setting root of OF to " << value;
        
        setOFRoot(value);
	}
    
    void handleProjectPath(const std::string& name, const std::string& value){
		
        ofLogNotice() << "setting project path to " << value;
        
        projectPath = value;
	}
    
    void handleAddons(const std::string& name, const std::string& value){
		
        addons = ofSplitString(value, ",", true, true);
    }
```

are never used

arturoc:
this is what i do in the old PG:

```cpp
projectPath = ofFilePath::removeTrailingSlash(ofFilePath::getPathForDirectory(ofFilePath::getAbsolutePath(arg,false)))
```

that works for all cases

ofZach:
handleOFPath and handleAddons are used if you use those flags, but handleProjectPath should be removed.  I'll take a look, thanks for the crazy ofFilePath trick, it feels like (http://www.matryoshka.biz/images/7/7011__2.jpg), looking into it.

arturoc:
mmh i don't see any call to handleAddons or handleOFPath anywhere in the code

ofZach:
sorry my bad, I was responding on instinct...  you're right, they're handled elsewhere.

ofZach:
ok I think this should work, together with this PR 
https://github.com/openframeworks/openFrameworks/pull/4246
I am able to relatively update using . or ../ as the parameter passed into the command line tool. 

I moved some stuff around to make sure the ordering was right in the command line tool, requires some testing.

In testing I noticed that all my xcode projects were broken based on double find and replaces, thus the PR against the main branch.  I'm not sure why I never noticed that before  ¯\_(ツ)_/¯

arturoc:
:) should we merge it then?

ofZach:
yes!  let's do it !!

ofZach:
( I do think the readme and help screen need some love but I will take a look at this later on a larger push to make the output of the PG much more user friendly  )

arturoc:
yay!

arturoc:
mmh, did you removed the scripts folder?

ofZach:
oh sorry, I didnt know what that was...  what is that, btw?

arturoc:
it's the scripts so the tests on travis compile the PG automatically when there's any change. that way we don't have to recompile it for the nightly builds or the release.

ofZach:
oh sorry !!  I'll re-add them...  I thought that was something I did (I have alot of random stuff around when I was trying to figure out how to extract the command line tool)

arturoc:
i just readded them no worries

PR #15, opened 2015-08-17
cleaning up logging

ofZach:
I am starting to clean up command line logging and feedback.  I will open a PR on OF as well with changes to the addon

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -57,7 +57,8 @@ public:
 	string              ofPathEnv;
 	string              currentWorkingDirectory;
 
-
+    
+    bool bAddonsPassedIn;
 	bool bVerbose;                          // be verbose
 	bool bForce;                            // force even if things like ofRoot seem wrong of if update folder looks wonky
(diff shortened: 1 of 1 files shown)


ofZach:
@Daandelange (cc @kylemcdonald) these updates (besides fixing alot of the spurious / bad output) allow you to do "update" of a existing project but specify the addons via parameter (so it takes that list instead of parsing addons.make).  This should be the way to proper updating via the gui.

ofZach:
added timing and summary at the end, ala

     1 project updated in 0.015625 seconds

ofZach:
I'm going to merge this, there may be some more work on this but since the main OF pr was merged, this makes sense to merge and it would help to not diverge too much

kylemcdonald:
cool! small suggestion, the duration (and its accuracy) might be a distraction rather than helpful for quick updates. maybe:

```c++
if(elapsedTime > 1) {
    cout << "in " << int(elapsedTime) << " seconds";
}
cout  << endl;
```

ofZach:
sure good point, I'll add that...

PR #21, opened 2015-08-24
Fix cwd

### LINKED ISSUE #18: commandLine issues
in osx, from OF rooot:

```
apps/projectGenerator/commandLine/bin/projectGenerator -o /Users/arturo/Desktop/openFrameworks/ examples/addons/allAddonsExample
```

tries to create the project in:

```
project path is: /Users/arturo/Desktop/openFrameworks/apps/projectGenerator/commandLine/bin/examples/addons/allAddonsExample
```

and:

```
apps/projectGenerator/commandLine/bin/projectGenerator -o . examples/addons/allAddonsExample
```

fails to parse `.` as a directory
apps/projectGenerator/commandLine/bin/projectGenerator -o /Users/arturo/Desktop/openFrameworks/ examples/addons/allAddonsExample
```

ofZach:
this should (I think) fix #18 -- specifically I moved from using path::current() to getenv("PWD") as path::current and cwd were returning the path to the binary not the current location where you were executing it.  I think when I used an alias to test, I didn't see the problem maybe?

(also, I need to check if this works on windows / I assume this works on linux)

I tested relative and absolute paths from the OF path and for the project to be generated, everything that's relative is relative from where you are in the terminal (not relative from the root of OF, etc).  

this invites testing & whenever I am dealing with relative paths wonkiness, I start to feel a little like: 

giphy

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -471,6 +471,38 @@ public:
 		}
 		if (!bDryRun && !bConsiderAddonsDotMake) project->save(true);
 	}
+    
+    
+    
+    std::string get_working_path()
+    {
+        char temp [ PATH_MAX ];
+        
+        if ( getcwd(temp, PATH_MAX) != 0)
+            return std::string ( temp );
+        
+        int error = errno;
+        
+        switch ( error ) {
+                // EINVAL can't happen - size argument > 0
+                
+                // PATH_MAX includes the terminating nul,
+                // so ERANGE should not be returned
+                
+            case EACCES:
+                throw std::runtime_error("Access denied");
+                
+            case ENOMEM:
+                // I'm not sure whether this can happen or not
+                throw std::runtime_error("Insufficient storage");
+                
+            default: {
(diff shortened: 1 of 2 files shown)


ofZach:
oof, it looks like a commit from another branch got in here (to fix the PG simple compilation... not sure if I need to unwind this)

arturoc:
i think it's fine, the commit wasn't merged yet but it's now so it shouldn't matter.

as i posted in the issue, `std::filesystem::current_path()` should work, also using `std::filesystem::canonical(somepath)` gives you the absolute path without . or .. which is quite handy to convert the input path to something valid. it'll throw an exception if the path doesn't exist though in which case you can use `std::filesystem::absolute(somepath)`

ofFilePath::getAbsolutePath(somepath)` does a combination of both to remove . and .. before converting to absolute path

arturoc:
this is fixed now with: https://github.com/openframeworks/openFrameworks/pull/4299#issuecomment-136419943

PR #22, opened 2015-08-24
different approach on the fix

ofZach:
@arturoc here's a fix that grabs the current working directory at the top of the app (before it get's mucked with via ofUtils) and it seems work.  This seems maybe more cross platform and simpler then the  other fix I put up.

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -60,6 +60,7 @@ public:
 	vector <int>        targets;
 	string              ofPathEnv;
 	string              currentWorkingDirectory;
+    
 
     bool bVerbose;
     bool bAddonsPassedIn;
@@ -74,6 +75,12 @@ public:
 
 	commandLineProjectGenerator() {
         
+        
+        // on OSX there's some wonkiness in ofUtils that messes with CWD on OSX.
+        // Let's store it here and use it later without worrying about that
+        
+        currentWorkingDirectory = Poco::Path::current();
+        
         // this is called before params have been parsed.
         
         bAddonsPassedIn = false;
@@ -86,6 +93,7 @@ public:
 		bHelpRequested = false;
 		targets.push_back(ofGetTargetPlatform());
 
+        
         
 	}
 
@@ -122,7 +130,7 @@ public:
 
 
 
-		currentWorkingDirectory = Poco::Path::current();
+		
 		loadConfiguration(); // load default configuration files, if present
 		Application::initialize(self);
 
@@ -489,7 +497,7 @@ public:
 
 		//-------------------------- get the path to the current working folder
 
-		Path cwd = Path::current();                      // get the current path
+		Path cwd = currentWorkingDirectory;                      // get the current path
 		projectPath = cwd.resolve(projectPath).toString();  // resolve projectPath vs that.
 		Path resolvedPath = Path(projectPath).absolute();         // use absolute version of this path
 		projectPath = resolvedPath.toString();
@@ -536,13 +544,7 @@ public:
 
 
 
-		if (!isGoodOFPath(ofPath)) {
-			consoleSpace();
-			ofLogError() << "path to openframeworks (" << ofPath << ") seems wrong, please check";
-			consoleSpace();
-			return Application::EXIT_OK;
-		}
-
+		
         
 
 		if (ofDirectory(projectPath).exists()) {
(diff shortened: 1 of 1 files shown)


arturoc:
this seems super hacky, i remember the problem with the data path but i don't see how just doing:

```cpp
string defaultDataPath(){
    return string(ofFilePath::join(ofFilePath::getCurrentExeDir(),  "../../../data/"));
}
```

should do anything different than the current solution, `getCurrentExeDir` uses the same `_NSGetExecutablePath` to get the exe path but instead of changing the current dir we would just be querying it and adding it to the `"../../../../data"` that we are adding now to the current dir.

it should be even more robust since with the current solution if some app internally changes the current dir using c or c++ calls it'll break `ofToDataPath`

ofZach:
cc @ofTheo since that OSX thing is from a long time ago (glut days). 

with that change ofUtils, ofToDataPath is fine, but I can get two different results for std::filesystem::current_path() when I use glut window.   This settings gets toggled automatically (I think it's based on if there are external files in your .app) and the CWD is fairly unpredictable: 

this is the settings, you can adjust it by hand as well: 
screen shot 2015-08-25 at 4 56 43 am

running the app two times, with this: 

    cout << std::filesystem::current_path() << endl;
    cout << ofToDataPath("blah") << endl;

I see: 

	"/Users/zachlieberman/Documents/openframeworks/git/openFrameworks/apps/myApps/545454/bin"
	/Users/zachlieberman/Documents/openframeworks/git/openFrameworks/apps/myApps/545454/bin/545454Debug.app/Contents/MacOS/../../../data/blah

the second time, if I modify the project (or adjust the settings) I get: 

	"/Users/zachlieberman/Documents/openframeworks/git/openFrameworks/apps/myApps/545454/bin/545454Debug.app/Contents/Resources"
	/Users/zachlieberman/Documents/openframeworks/git/openFrameworks/apps/myApps/545454/bin/545454Debug.app/Contents/MacOS/../../../data/blah

so this is giving a CWD that's pointing to resources, not where the .exe is even (in MacOS). 

I think some of what's happening in that ofSetWorkingDirectoryToDefault is trying to standardize the CWD results, at least in glut windows on OSX.

arturoc:
I guess if glut changes the current path and it's a documented thing, as hacky as it is, that's fine, i guess it's a feature in the sense that, that way you can load resources from inside the bundle without specifying a path. But as long as ofToDataPath keeps working we shouldn't touch the current path.

But with the current fix, glfw, which doesn't touch the cwd, would be broken, i would expect the current directory to be the directory from which i've run the app, mostly for command line applications. Also if someone does `setcwd(other_path)` or `std::filesystem::current_path(other_path)` ofToDataPath would stop working.

With the fix i proposed above, the app would still be able to find `data`  no matter where it's run from and will keep the correct cwd.

Also `ofFilePath::getCurrentWorkingDirectory()` is reporting cwd as the exe directory which is not correct either, if someone needs to know the exe path they can use `ofFilePath::getCurrentExeDir()`.

ofZach:
I think this fix (which is hacky btw) would work with everything (since it's grabbing CWD at the start, and it's going to be unaltered by the ofUtils code at that point) but your solution would also work, but potentially break projects that expect CWD to be consistent on glut.  As glut is on it's way out, maybe it's ok and we just document that people should look to ofFilePath::getCurrentExeDir() for stability with regards to knowing where the app is.   I don't mind either approach, I feel like I'd like to hear from @openframeworks/macos / @ofTheo on this just to make sure changing ofUtils is ok.   

We could also move the weirdness of ofUtils setting the CWD to ofGlutWindow so that it's not applied across the board on OSX but just in the case of opening up a glut window.

arturoc:
i think even in the glut case this is better than what we have now, glut is changing the cwd which is documented and there's even an option to change that behaviour in it's menus, but right now even that option won't work because we are overriding it always.

this behaviour is not documented and it's not very standard.

my only worry with your solution, even if it works now, is that if someone in the future modifies this and uses the cwd instead of the variable you've added it's going to introduce a bug that is really difficult to detect.

i don't mind merging this as it is as it'll keep working with my fix too but i would then merge this: https://github.com/openframeworks/openFrameworks/pull/4299 too, unless @ofTheo has any insights on why it might fail

ofZach:
sure your solution works with ofToDataPath which was the big thing we were trying to fix with this -- it was terrible when glut apps randomly couldn't load data b/c of the cwd change :)  I still feel it's weird when glut apps change the cwd, but I can live with that and since glut is on its way out, it feels ok.   Maybe we can even add a convenience function to ofAppGlutWindow (ie, ofAppGlutWindow::standardizeCurrentWorkingDirectory)  which does this and explains why you might want to normalize the CWD so that things like Path::Current() are consistent.   I feel like having it as an optional function with some comments might be enough (rather then doing it automatically)

arturoc:
the problem is that the exe dir is not always the standard cwd, instead of using the current fix we could query the cwd in the glut window before calling glutInit() and store it to use it in a potential `standardizeCWD`

ofZach:
that's a good option -- I just remember alot of pain with cwd changing, I think that ofToDataPath works fine mitigates that, but having some sort of solution to make sure things don't change behind your back is really useful.

arturoc:
this is fixed now with https://github.com/openframeworks/openFrameworks/pull/4299#issuecomment-136419943

PR #25, opened 2015-08-28
Add LibraryBinary files to pgLegacy VS project

### LINKED ISSUE #24: pgLegacy won't compile on VS2015
pgSimple compiles and starts (I didn't test any actual functionality) without issue, but I get linking errors related to LibraryBinary for pgLegacy. I think whatever happened this commit https://github.com/openframeworks/projectGenerator/commit/3e412abe22c5a8030f909a65676c57433eacbfb0 needs to happen as well? That looks like a non-trivial reorganization so I don't want to muck things up...

The reason I'm bothering with pgLegacy rather than using the commandLine tool is that I can't add addons that are outside of OF_ROOT\addons, which I reported here https://github.com/openframeworks/projectGenerator/issues/23

mattfelsen:
Fixes #24


arturoc:
thanks!

PR #26, opened 2015-08-29
Support for local addons

arturoc:
Adds support for local addons through the command line or addons.make

Now an addon can be specified with a path relative to the project path
and it'll be included in the project using that path.

So addons.make can now be something like:

```
ofxGui
../addons/ofxSomeLocalAddon
```

which will include ofxGui from OF/addons/ofxGui and ofxSomeLocalAddon
from the folder addons/ofxSomeLocalAddon one level above the project

Similarly the command line PG can be run like:

```bash
projectGenerator -a "ofxGui, ../addons/ofxSomeLocalAddon" newProject
```

which will create a new project adding the addons specified in the same
way explained for addons.make

This also brings back the common code in ofxProjectGenerator which
was in the core repo to the PG repo, which should make maintainance easier


ofZach:
exciting ! can't wait to test this (already thinking of how to expose this in the gui version).  
cc @mattfelsen

ofZach:
ps: the changes to  commandLine/Project.xcconfig which overwrite my changes seem wrong, I need to remove fmod (and a few core OF libs) to make the standalone work (otherwise, there's dylib issues) so I've done that in this hand edited config...   I guess it's been overwritten by a template?

arturoc:
oh, ok yes i've recreated all the projects using the PG, we should have some way to do this automatically so we can regenerate the PG projects with the PG itself. i'm going to take a look

ofZach:
thanks!  yes this was a very specific hack since if you link against the framework there is weird references to the dylib you can't remove, which makes it really hard to build OSX command line apps.  I tried a variety of approaches and tools to remove the linkages but no dice, so the easiest was to edit the config directly. 

I also have a post build step in the xcproj (which is likely clobbered by this commit) which rips the command line app out of the .app, deletes the .app and renames the command line tool.  It's kind of useful to have on OSX for building the command line tool.

ofZach:
that line is here btw: 

https://github.com/openframeworks/projectGenerator/blob/master/commandLine/commandLine.xcodeproj/project.pbxproj#L302

it says: 

	cp "$TARGET_BUILD_DIR/$PRODUCT_NAME.app/Contents/MacOS/$PRODUCT_NAME" "$TARGET_BUILD_DIR/commandLinePG"
	rm -rf "$TARGET_BUILD_DIR/$PRODUCT_NAME.app"
	mv "$TARGET_BUILD_DIR/commandLinePG" "$TARGET_BUILD_DIR/projectGenerator"

I'm not exactly sure how to let the command line PG built itself on OSX, it's a little bit specific :)

arturoc:
i guess we can have some flag in config.make. for the makefiles config.make it's already setting the name of the binary to projectGenerator instead of commandLine. we could add a specific flag for osx post steps that would be run after the app is built. this would be easy to parse from the makefiles and the PG

mattfelsen:
Oh this sounds awesome – I'm sure this would make @quinkennedy happy too :smile: Thanks! I'll give this a whirl on Windows soon

ofZach:
I still don't know how we don't alter the xconfig -- would that be a config.make setting?  I can look at how we add post build steps, but I'm a little worried about implementing them across all platforms in the PG right now.  Do you think for 0.9 we can use the hand modified xconfig and xcproj files and then add these features after 0.9 ?

arturoc:
it's just a pain to have to open every platform to create the project each time we change something.

a pretty easy way to do it would be to have specific templates with different features. they won't have everything so we don't have to maintain multiple copies of the project files but we would copy first the original and then overwrite whatever files there's in the specialized template.

this would also be useful to add templates for using the programmable renderer for example, or have more than one window...

mattfelsen:
I'm getting compile errors for ofRestoreWorkingDirectoryToDefault()...I see in the commit that changed from Set to Restore, but I don't eitherfunction defined anywhere.

ofZach:
I think templates would be great to look at post 0.9.   

I am going to look at adding post build step to the project generator now / config.make since that seems useful, but I'm still not sure how we deal with the config file.  Maybe there could be an exclude files for when you copy over the template?   I don't know how useful that is.   

Another option is a post generate step (ie, an instruction that gets called after the project gets generated).  In this case, we could replace out the config file.

either way, I'm also down to solve this after 0.9, my head is kind of in the gui and this seems useful but not crucial (compared to getting frameworks to work on osx, etc). 

another option is to make certain new features we are discussing mac specific for now, then add them to all platforms later (b/c I can test on osx and it's harder to add it across all project types)

arturoc:
hey, i have this almost working, i'm also formatting the output better, let me a moment to finish it and upload all i have

ofZach:
:tada::tada::tada::tada::tada: cool !!

mattfelsen:
I misspoke a bit last time...I do in fact have `ofSetWorkingDirectoryToDefault()` and changing Restore back to Set compiles and works for me. I'm confused why this would pass on Travis though...am I the only one seeing this?

arturoc:
travis is only compiling the PG simple right now. 

That's failing because i'm using a branch for the OF core with a different fix. in any case that call is not necessary and the correct one is ofSetWorkingDirectoryToDefault()

mattfelsen:
:+1:

arturoc:
what if we change `-x` to  `-p` to specify the platform? we are using `-x` because `-p` was used to specify the path but now that the path is an argument and not an option i thikn it makes more sense

ofZach:
sounds good, the x seemed cool at the time 

x-factor-flash-vote

but p seems more reasonable...

arturoc:
this also includes custom templates now, once https://github.com/openframeworks/openFrameworks/pull/4321 is merged. now the command line tool can be called like:

```bash
projectGenerator -l -p osx
```

and it will list the available templates for that platform (or the current platform if non is passed) for osx for example this returns:

```
standard
nofmod
gl3.2
gl3.3
gl4.0
gl4.1
gl4.2
gl4.3
gl4.4
gl4.5
```

gl4.3+ aren't really supported by osx but we can just remove them from the downloadable package instead of duplicating them for every platform

calling

```bash
projectGenerator -t gl3.2 projectName
```

creates a project which creates the window using openGL 3.2 in main.

ofZach:
this is cool!  checking it out now.  just a style thing but local_addons might be better as camel case if we aiming for consistency:

screen shot 2015-09-01 at 4 30 04 pm

arturoc:
makes sense, feel free to change it, it's only in one or two places in `xcodeProject.cpp`

ofZach:
yep found it ! thanks...

PR #28, opened 2015-09-01
adding osx info about how to make an executable and add a shell script

ofZach:
this is adding a shell script for pulling the executable out of the .app and added some info in the readme (this used to be part of the xcproj build step but is now broken out since we've lost that using the templates)

### DIFF
--- a/commandLine/prepareOsx.sh
+++ b/commandLine/prepareOsx.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+echo "copying the executable out of the app"
+
+cp bin/commandLine.app/Contents/MacOS/commandLine bin/commandLineExe
+
+echo "deleting the app"
+
+rm -rf bin/commandLine.app
+
+echo "renaming the executable"
+
+mv bin/commandLineExe bin/commandLine
+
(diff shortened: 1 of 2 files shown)


arturoc:
oh, i thought this was working but totally forgot to check it.

this line in config.make:

https://github.com/openframeworks/projectGenerator/blob/master/commandLine/config.make#L30

adds a rule to execute after compiling, it should work both on the makefiles and the generated xcode project. in the project it's here:

https://github.com/openframeworks/projectGenerator/blob/master/commandLine/commandLine.xcodeproj/project.pbxproj#L440-L458

but it might be that the xpath rule is adding it the wrong place?

ofZach:
ah sorry !  I just looked at the project and thought it was missing.  it looks like it's not made it all the way through in the xcproj xml (there's sometimes one or two things you need to do get new things registered in the right way)...  let me take a look and see if I can fix it up.

ofZach:
closing this PR, gonna work on fixing the build step :)

arturoc:
great, thanks!

ofZach:
oh it works!  PR coming shortly :)  this is pretty cool !!   also, I can tell when you've written the code because I see the word auto and comments like: 

cout << "begin" << ofTrimBack(" hola ") << "end" << endl;

PR #29, opened 2015-09-02
Osx fix2

ofZach:
not ready to be merged yet -- but fixes the after rule for osx.  some notes: 

- should we call "after rule"  "post build script" (just a thought) 
- there are two buildPhase UUIDs, one is for the actual building one is to access the array of steps, look into the naming.  Maybe look into renaming these so they are more clear. 
- check IOS for UUID similarities or differences, test on IOS. 
- clean up some of the couts around parsing the configs (make it notice / verbose) for logging


ofZach:
ah I see that I have commits from the previous osxFix branch in here, let me remove those...

ofZach:
this fix seems good now, I regenerated the command line project using the command line app and the nofmod template and it compiles, etc.  removed the unnecessary .sh script and cleaned up the readme (we need more info about usage in there though).   ping @armadillu this should compile for you now...

ofZach:
also, are we cool with the term "after rule"  ? -- I generally prefer "post build script" but after rule seems ok...

ofZach:
I'm going to merge this since osx command line is fairly broken until this is....   I'm trying to add more readme notes now but will do that against master.

PR #49, opened 2015-10-05
added shrinkwrap.json to fix downstream dependencies

ofZach:
this helps fix a downstream dependency that needs to be updated for 10.8 on OSX to work (as reported by @nickhardeman).    We need to use this approach until electron packager's dependencies get updated.

to fix the dependencies, you will need to refresh the install: 

    for package in `ls node_modules`; do npm uninstall $package; done;
    npm install

### DIFF
--- a/projectGeneratorElectron/npm-shrinkwrap.json
+++ b/projectGeneratorElectron/npm-shrinkwrap.json
@@ -1,0 +1,1785 @@
+{
+  "name": "projectGenerator",
+  "version": "0.0.1",
+  "dependencies": {
+    "electron-debug": {
+      "version": "0.2.1",
+      "from": "electron-debug@>=0.2.1 <0.3.0",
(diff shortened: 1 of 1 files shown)


arturoc:
should we merge it? the create_package script downloads the repository every time and buillds from there so it shouldn't need to refresh the install right?

ofZach:
if the script redownloads all the node packages every time before we should be fine, but if it's cached, we need to run the above commands to update the packages.  This fixes a bug in the electron packager that was unzipping files incorrectly I think missing empty folders, etc.   The lack of the right folder structure was causing the OSX project to fail on 10.8.

it's probably a good idea to keep an open issue on this since I think it's ideal not to use shrinkwrap if at all possible (it's very explicitly listing dependencies so you can hand edit them) and as soon as this is fixed in electron packager, we can move back to just using the simpler package.json...

arturoc:
ok, i'm going to merge it so tomorrows nightlies have the patch and we can test if it fixes 10.8

Issue #52, opened 2015-10-10
abort() called on generate project


crucialfelix:
The Electron app alerts me:

```
Error...
There was a problem generating your project... /Users/crucial/code/idmx/of_v20151008_osx_release/apps/myApps/PlaySPLOM

Command failed: /bin/sh -c /Users/crucial/code/idmx/of_v20151008_osx_release/projectGenerator-osx/projectGenerator.app/Contents/Resources/app/app/projectGenerator  -o"/Users/crucial/code/idmx/of_v20151008_osx_release" -a"ofxGui,ofxOsc" -p"osx" /Users/crucial/code/idmx/of_v20151008_osx_release/apps/myApps/PlaySPLOM
```

OS X crash reporter reports:

```
VM Regions Near 0:
--> 
    __TEXT                 00000001032a8000-000000010398b000 [ 7052K] r-x/rwx SM=COW  /Users/USER/*/projectGenerator.app/Contents/Resources/app/app/projectGenerator

Application Specific Information:
abort() called

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff94576286 __pthread_kill + 10
1   libsystem_c.dylib             	0x00007fff88eba9b3 abort + 129
2   projectGenerator              	0x000000010337b15d Poco::SignalHandler::handleSignal(int) + 45
3   libsystem_platform.dylib      	0x00007fff8e7d5f1a _sigtramp + 26
4   projectGenerator              	0x0000000103533ec6 std::__1::__tree_node_base<void*>*& std::__1::__tree<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::less<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >, std::__1::allocator<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > > >::__find_equal<std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >(std::__1::__tree_node_base<void*>*&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 38
[...]

Issue #51, opened 2015-10-09
Abort trap: 6


robert-wallis:
I'm getting a strange runtime error with both github and the nightly:

```
$ /Users/robert/Downloads/of_v20151008_ios_release/projectGenerator-ios/projectGenerator.app/Contents/Resources/app/app/projectGenerator -r  -o"/Users/robert/3rdParty/openFrameworks" -p"ios" /Users/robert/3rdParty/openFrameworks/examples/ios
Abort trap: 6
```

XCode 7.0.1, OS X 10.11

When debugging the commandLine project the error happens in the `POCO_MAIN_APP(commandLineProjectGenerator)` macro.

The stack trace is stopped on 0x1000d3c68 in `Poco::Util::Application::init()`

```
    0x1000d3c60 <+608>:  movq   %r14, %rdi
    0x1000d3c63 <+611>:  callq  0x1000d4060               ; Poco::Util::Application::processOptions()
    0x1000d3c68 <+616>:  leaq   -0x98(%rbp), %rdi
    0x1000d3c6f <+623>:  callq  0x100144ff0               ; Poco::Path::~Path()
    0x1000d3c74 <+628>:  addq   $0x258, %rsp
```

Any help with this would be great, I am unable to generate iOS projects, but OS X projects generate just fine.

ofZach:
I've discovered that it has something to do with parsing the command line options and differences between these two characters: 

— 
vs
-

for example this fails: 

./projectGenerator -o "../../../../" blah

but this works: 

./projectGenerator —o "../../../../" blah

unfortunately this "-" is what you get when you type a minus symbol, the other symbol I got from copying and pasting.

I'm debugging the crash and I see this in the stack trace, which makes me think it's about parsing: 

```
[...]

PR #53, opened 2015-10-12
new options parser

ofZach:
this removes poco for the command line options because of the crashing that's been reported: 

https://github.com/openframeworks/projectGenerator/issues/52
https://github.com/openframeworks/projectGenerator/issues/51

and uses another open source options parser that just a header file.   

It's not 100% working yet (I am seeing some issues on projects that I make on osx with this) but opening this PR work through this change.


ofZach:
addons are working now -- you need to specify without spaces or use the longer form: 

     ./projectGenerator -a"ofxXmlSettings" myProject
     ./projectGenerator --addons=ofxXmlSettings myProject

ofZach:
OSX seems for the most part working now (I think help screen can use love and examples), going to test on windows...

ofZach:
windows seems good, tested both 64 and win32 versions and it generated ok.   cc @mattfelsen if you want to test this...   

quick notes: 

- The project path it's calculating looks mangled sometimes (but works), would be good to fix paths internally so you don't see as much half-posix half windows paths. 
- I need to look at the template variable and what it should be set to by default

robert-wallis:
It's working!  It's working!
It's alive

ofZach:
@arturoc can you test this with linux?  On windows / mac seems to be good.  If it's ok, I will update the electron project and we can test this.  Also ping @mattfelsen if you want to test windows

ofZach:
@arturoc also, I see a print out: 
```
[notice ] using additional template
```
but template is just set to "" -- not sure what this print out means...  I'm trying to clean up some of the output...

arturoc:
i'll give it a try tomorrow, but if you want to merge it already go for it, that way we can check if the package generation in the server works which will test linux too  :)

```
[notice ] using additional template
```

this is used when you pass for example:

```
projectGenerator -t gl3.3 myProgrammableRendererApp
```

which will copy the files in that template to the project apart from the platform ones. not sure what's the problem but it should only show if there's really a template.

mattfelsen:
I must be cursed.

Here's what I'm seeing...

Mode | Task | Notes
--- | --- | ---
Release | Create project | ok
Debug| Create project | ok
Release | Update project | crash
Debug| Update project | ok
Release | Create project with addons | ok, but no addons added to project
Debug| Create project with addons | ok, but no addons added to project

I've got my %PG_OF_PATH% set and added the path to the commandLin binaries to my %PATH%. I tried all of these syntaxes:

- `--addons="ofxGui"`
- `--addons=ofxGui`
- `-a"ofxGui"`
- `-a="ofxGui"` just for kicks

```
C:\Users\mattfelsen\Code\openFrameworks\github\apps\myApps>commandLine_debug.exe -v CreateAddonsD --addons=ofxGui

[notice ] setting OF path to: C:\Users\mattfelsen\Code\openFrameworks\github
[notice ] from PG_OF_PATH environment variable
[notice ] project path is: C:\Users\mattfelsen\Code\openFrameworks\github\apps\myApps\CreateAddonsD
[notice ] using additional template
[notice ] setting up new project C:\Users\mattfelsen\Code\openFrameworks\github\apps\myApps\CreateAddonsD
[verbose] loaded filter
[verbose] returning path ../../../

[notice ] saving addons.make
[notice ] project created!


1 project created [notice ] in 0.613159 seconds



C:\Users\mattfelsen\Code\openFrameworks\github\apps\myApps>
````

Gonna dig in and see if I can figure out why the Release mode crash is happening, and if the addons thing happens if I specify -o/-ofPath on the command line instead of from the env var

**Comment by ofZach** [comment]
yeah ! there's some voodoo....

I don't think you can specify the options in this order: 

```commandLine_debug.exe -v CreateAddonsD --addons=ofxGui```

can you try: 

```commandLine_debug.exe -v --addons=ofxGui CreateAddonsD```

(this new system is more specific about ordering)

**Comment by mattfelsen** [comment]
!! Wow, can't believe I didn't notice or try that. Putting addons before the project path works, and actually, it doesn't seem like putting it after should be supported -- I guess I just got used to it working before. Looking at the crashes now

**Comment by ofZach** [comment]
yeah the poco was pretty forgiving about order, this one less so, but it's on the other hand it's somewhat nicer in terms of it's output to the console for help, and code wise it's much smaller.  Also, it's great to have windows and posix commands be the same....

**Comment by mattfelsen** [comment]
Whew, ok, noticed that it was crashing on the same line as before (generateUUID). I added the same hack (returning a copy of the digestString) and no more crashes :smile:

**Comment by ofZach** [comment]
ok thanks !  I'll add that hack ...

**Comment by ofZach** [comment]
@mattfelsen is this what you mean?
https://github.com/openframeworks/projectGenerator/commit/3339b4f79451ca4afe9fe3ba41e9cdfa9df0716d

I'm cleaned up electron and did some testing with this.  Also added a nice topbar based on @julapy 's awesome image from flickr https://www.flickr.com/photos/julapy/16796836033/in/pool-openframeworks/ 
(@julapy hope that's ok ?!) 

screen shot 2015-10-14 at 12 52 32 am

**Comment by ofZach** [comment]
@arturoc merged this in, fingers crossed....

**Comment by julapy** [comment]
@ofZach just noticed your comment about the flickr image.
totally cool!

**Comment by arturoc** [comment]
:) great, there was a problem with linux but it's solved now and also the empty templates message.

the merge it was too late to get into today nightlies but tomorrow's will have the new command line parser.

i've also merged the qtcreator changes and changed both linux and mingw to have qtcreator as the default project instead of codeblocks.

**Comment by mattfelsen** [comment]
@ofZach that's the one :)

**Comment by ofZach** [comment]
thanks!  saw  it in the merge.   It would be good to test local addons on windows as well as batch updating -- I think they should be ok since not much has changed in the parsing...

I am curious about the gui as well on windows, I removed the windows specific calls and since OSX works I imagine windows would work as well.  I think there's some small fixes needed for things like paths with spaces in them, etc.

**Comment by ofZach** [comment]
@arturoc cool !  I figured nightly might mean something different in berlin vs nyc...   

I feel better about this command line tool, I am going to work through some of the outstanding issues with the gui now (quotes around the project path, etc).   it will be good to see that the 10.8 osx issue is solved also.

**Comment by mattfelsen** [comment]
@ofZach Good call, I can test that stuff on Windows tonight. Are there any special build instructions for the gui on Windows other than what's in the readme?

**Comment by ofZach** [comment]
this sparse info should probably be enough to get started

https://github.com/openframeworks/projectGenerator/tree/master/projectGeneratorElectron

but I imagine you also need to alter the exe that the command line makes (rename it projectGenerator and put it and the dlls here): 

https://github.com/openframeworks/projectGenerator/tree/master/projectGeneratorElectron/app

```npm start``` will fire it up.   Also, you can run a specific build process to package it up:  

```npm run build:vs```

to make just the electron for your platform

I'll take a look also.  I just installed window 10 on a machine and I'm finding it significantly easier to test and jam on things.  Windows 8 was driving me crazy.... 

thanks!

**Comment by mattfelsen** [comment]
Local addons look like they're still working, but with one caveat. The path to the local addon used to be relative to the project folder, but now it seems like the pg is looking for it relative to the $OF_ROOT/addons folder.

For example, I've got this structure...
```
$PROJECT_ROOT
- openFrameworks
-- addons
-- apps...
- ExternalAddons
- Sketches
```

This used to work from inside the Sketches folder
`commandLine.exe -a"../ExternalAddons/ofxCv" newProject`

But now I have to use
`commandLine.exe -a"../../ExternalAddons/ofxCv" newProject`

Here's some (truncated) output to show the paths the pg is searching for (this does not pick up the addon)

```
C:\Users\mattfelsen\Code\Projects\CMA2\Sketches>commandLine.exe -v --addons="../ExternalAddons/ofxJSON" LocalAddonTestR


[verbose] listed 8 files in "C:\Users\mattfelsen\Code\Projects\CMA2\Sketches\LocalAddonTestR\"
[notice ] setting OF path to: C:\Users\mattfelsen\Code\Projects\CMA2\openFrameworks
[notice ] from PG_OF_PATH environment variable
[verbose] in fromFS, trying src C:\Users\mattfelsen\Code\Projects\CMA2\openFrameworks\addons\../ExternalAddons/ofxJSON/src
```

(this does pick up the addon)
```
C:\Users\mattfelsen\Code\Projects\CMA2\Sketches>commandLine.exe -v --addons="../../ExternalAddons/ofxJSON" LocalAddonTestR

[verbose] in fromFS, trying src C:/Users\mattfelsen\Code\Projects\CMA2\Sketches\LocalAddonTestR\../../ExternalAddons/ofxJSON/src
[verbose]       ofxJSON.h
[verbose]       ofxJSONElement.cpp
[verbose]       ofxJSONElement.h
[verbose] listed 3 files in "C:/Users\mattfelsen\Code\Projects\CMA2\Sketches\LocalAddonTestR\../../ExternalAddons/ofxJSON/src\"
```

Nice that using either / or \ in the paths works on Windows though :grinning:

**Comment by arturoc** [comment]
This new parser is kind of weird, now you can't do:

```
projectGenerator -o path -p platform projectPath
```

but instead:


```
projectGenerator -opath -pplatform projectPath
```

which is not very standard. let's leave it like this for now but would be good to find another solution at some point.

**Comment by arturoc** [comment]
@mattfelsen what platform are you in? this seems to be working for me in linux and windows

@ofZach it seems the new command line parser is failing in the server:

```
terminate called after throwing an instance of 'std::regex_error'
  what():  regex_error
./create_package.sh: line 131: 16429 Aborted                 ${main_ofroot}/apps/projectGenerator/commandLine/bin/projectGenerator --recursive -p${pkg_platform} -o$pkg_ofroot $pkg_ofroot/examples > /dev/null
Error on or near line 140; exiting with status 1
````

if the library is using std::regex it won't work in gcc < 4.9. i'm going to look if i can update the server's gcc somhow

arturoc:
ok, the problem is in the qtcreator project creation class not in the command  line parser

ofZach:
yeah you can't do this: 

projectGenerator -o path -p platform projectPath

but can do this: 

projectGenerator -o"path" -p"platform" projectPath

or 

projectGenerator --ofPath=path --platform=platform projectPath

I'm happy to use another parser, the requirement for 4.7 gcc meant I needed to try something a bit simpler.  I also liked that makerbot uses it https://github.com/makerbot/optionparser/tree/master/src.

ofZach:
ps happy to try Getopt at some point.  http://www.gnu.org/software/libc/manual/html_node/Example-of-Getopt.html#Example-of-Getopt

mattfelsen:
Ah, you know what, the ../../ **does** make sense because the path is relative to **inside** the project folder (i.e. `Sketches/newProject`), not the CWD from where I was running the pg (i.e. `Sketches/`). False alarm. :grimacing:

Issue #42, opened 2015-09-12
always report project generator string that's used in the console output


ofZach:
it's really useful to see the command line string that we are using in the console output so we can go test it by hand if need be, ie the "projectGenerator -o="....." -r ......"  string.    I think that used to be in there at some point.

Daandelange:
yes it was in the console, looks like it's been stripped out. Shall I put that back ?

ofZach:
yeah I think it's a good idea to put it back in, so we can debug easier any problems.

Daandelange:
There you go, with a readme entry and I also put back my last forgotten commit from ofZach/pgElectron

ofZach:
thanks !   I don't think we need it in an ever present console -- just to be reported on the console you see with success / failure dialog...

Daandelange:
It's pretty discreet like this and I find it useful useful to trace back generated projects and, as you said, look at their command line string. What about simplifying/shortening it so that you only see the command history, clicking on each of them them will show their console output.

The verbose mode setting already defines if we see the success modal with a console (btw, this setting does nothing for the moment... and is not present in the default `settings.json`)

So I see the verbose option more as a project related setting while the showConsole setting is related to all projects.

ofZach:
no I'm sorry we already decided not to have this console.  we already have 3 potential console outputs

a) the console output if you run electron from the command line (ie npm start) 
b) the console output if you run developer tools
c) the console output from success / failure on generate project. 

I really don't think we need a 4th and I think this just confuses things to have this user facing.  all I am asking for here is to always put the command line string that we are firing off the command line tool in the console output of (c).    We could improve the reporting in (a / b) if it feels like we need to give better feedback.   thanks!

ofZach:
(ps this is where we decided not to have this additional console https://github.com/ofZach/pgElectron/issues/24#issuecomment-134220973)

Daandelange:
Ok, that sounds good! Sorry for the confusion...
In fact i saw the showConsole as a hidden setting but by cleaning up the other consoles it's enough too. :+1: 

I added some commits removing the old console code and adding support for verbose option.

ofZach:
I've added this now -- closing this issue

PR #55, opened 2015-10-15
UI improvements/fixes + console removal

Daandelange:
See https://github.com/openframeworks/projectGenerator/issues/42


ofZach:
cool ! I think there's probably a more elegant solution for the launch / launching (ie, the button changes size and looks a little odd as it does so) but beyond that it looks good....   I will poke around semantic-ui and see if there's a way to give the button a loading feel, without it jumping in size...

Daandelange:
Sounds good :+1: 
Sorry for this kinda dirty PR but it should help :)

PR #57, opened 2015-10-17
allow for local addons to be specified via absolute paths

ofZach:
this is a PR to work through getting local addons to work in the gui.  

I am allowing them to be specified at the command line via absolute paths which get converted to relative paths to the project in the command line tool as the project is generated (ie, if you have a folder called localAddons in the root, and you pass in -a"/users/......./ofxTemp" it will be like adding ../../../localAddons/ofxTemp to your project) .  On the gui side, it's much easier to take a relative path to OF for the local addon (or absoute path) and store it internally as an absolute path....

I've added support in settings.json and in the gui now, just working on storing and passing through the right things, and on dealing with parsing addons.make on import. 

as I get stuff to work, I'll check it in...

### DIFF
--- a/ofxProjectGenerator/src/projects/baseProject.cpp
+++ b/ofxProjectGenerator/src/projects/baseProject.cpp
@@ -234,8 +234,18 @@ void baseProject::addAddon(std::string addonName){
     ofAddon addon;
     addon.pathToOF = getOFRelPath(projectDir);
     addon.pathToProject = ofFilePath::getAbsolutePath(projectDir);
+    
+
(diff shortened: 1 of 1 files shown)


arturoc:
that makes a lot of sense, specifying paths from the command line as relative to the project was kind of strange. only thing, be careful that specifying things through addons.make still works as relative from the project

ofZach:
going to merge this even if we not using these absolute paths via the gui....  I feel like it's generally a useful thing to have as an option.

Issue #58, opened 2015-10-17
AssetsLibrary added to iOS projects that require it


danoli3:
https://github.com/openframeworks/openFrameworks/issues/4290

Just for projects that require ```ofxOpenCv``` in their addons.make

So currently:
- opencvFaceExample
- opencvExample

ofZach:
I really think the easiest way to do this is via addon_config.mk -- see here for example: 

https://github.com/danomatika/ofxMidi/blob/master/addon_config.mk#L157

how ofxMidi adds the CoreMIDI framework on ios.

danoli3:
That seems to easy! I'll do it now

ofZach:
I think this is fixed now via this: 

https://github.com/openframeworks/projectGenerator/commit/3852fb414cca49198b326e21deab4992ac571657

Am I right to assume that this path: 

/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/

is the canonical path for system frameworks on ios ?

also, is this framework called "AssetsLibrary" ?  (I couldn't find AssetsFramework...)

now if the addon_config.mk looks like this: 

<img width="769" alt="screen shot 2015-10-18 at 5 59 58 pm" src="https://cloud.githubusercontent.com/assets/142897/10566917/4028ebfa-75c2-11e5-8e81-6c3880b59506.png">

you should see this in the output: 

<img width="291" alt="screen shot 2015-10-18 at 5 59 27 pm" src="https://cloud.githubusercontent.com/assets/142897/10566918/46fb693a-75c2-11e5-9d50-1df53191d7d7.png">

let me know if this helps....

ofZach:
ps: if you rebuild the command line tool to test this, this is the command I used: 

```./projectGenerator -p"ios" -o"../../../../" ../../../../examples/ios/opencvExample/
open ../../../../examples/ios/opencvExample/opencvExample.xcodeproj```

danoli3:
You are correct! It's AssetsLibrary D:

danoli3:
Okay now the issue really is with the project generator.

It adds the Framework to Xcode, however does not add the target to the project. 
So it still errors:

<img width="299" alt="screen shot 2015-10-19 at 9 35 28 am" src="https://cloud.githubusercontent.com/assets/830748/10567082/f5b65dd6-7644-11e5-85d2-54b4ee6c0f7e.png">

[...]

PR #64, opened 2015-10-18
fix for ios frameworks, need to double check osx system and non system

ofZach:
this is a fix for https://github.com/openframeworks/projectGenerator/issues/58 

I need to double check non ios system frameworks and regular frameworks, but this should work for IOS now....  

screen shot 2015-10-18 at 7 30 49 pm


ofZach:
osx tested with ofxMidi and ofxEdsdk and seems good, going to merge this.

PR #70, opened 2015-10-26
fixes for framework linking

ofZach:
this is the fix for https://github.com/openframeworks/openFrameworks/issues/3774 as reported by @bangnoise against the 0.9rc2


ofZach:
ok I am merging this now....  I think this fixes things but it should be tested.

PR #72, opened 2015-10-28
skipping some addons

ofZach:
this skips listing some addons in the PG list, which should help w/ errors like including ofxiOS twice.

### DIFF
--- a/projectGeneratorElectron/index.js
+++ b/projectGeneratorElectron/index.js
@@ -301,25 +301,9 @@ function parseAddonsAndUpdateSelect(arg) {
     //path = require('path').resolve(__dirname, defaultOfPath + "/addons");
     addons = getDirectories(arg + "/addons","ofx");
 
-    var a = addons;
-    var b = addonsToSkip;
-    var addonsThatShouldBeShown = [];
-    var found;
-    for (var i = 0; i < a.length; i++) {
-        found = false;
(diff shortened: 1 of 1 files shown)


projectGeneratorElectron/index.js
```diff
@@ -289,6 +300,27 @@ function parseAddonsAndUpdateSelect(arg) {
...
+                break;
+            }
+        }
+        if (!found) {
+            addonsThatShouldBeShown.push(a[i]);
+        }
+    }
+
+    addons = addonsThatShouldBeShown;
+
```
arturoc: this should work but in javascript you have this nice functional operators for collections where you can do something like:

```js
addons = addons.filter( function(addon) {
    return !addonsToSkip.contains(addon);
});
```

ofZach: cool -- but I get "addonsToSkip.contains is not a function" error....

arturoc: oh, might be a later addition, i think `includes` instead of contains should work and this should work for sure:

```js
addons = addons.filter( function(addon) {
    return addonsToSkip.indexOf(addon)==-1;
});
```

ofZach:
updated the filtering via @arturoc's comments....   this change seems good, I'm going to merge.

Issue #46, opened 2015-10-02
Feature request: open in xcode or any other ide button


thomasgeissl:
I would like to be able to open a successfully generated project in xcode or any other ide via the projectGenerator.
There could be an open button next to the log window's close button.
Maybe the following commands are useful:

osx: ``òpen $DESTINATION_PATH```

linux: ```xdg-open $DESTINATION_PATH```

kylemcdonald:
i like your command line solution! some other thoughts on this issue at #40

Daandelange:
I made this XCode feature in this commit, and it works great!
It should not be hard to add the linux method.
https://github.com/Daandelange/projectGenerator/commit/a17476e5a81f377cf98094cdc7eb622a7dd41d3b

ofZach:
I suggest we do this in the success modal for now, and then later versions drop the modal as @kylemcdonald suggests.  Right now, doing it w/ the update / generate buttons feels a little strange, the UI is kind of popping around, and I feel more comfortable adding it the modal then working it into the main ui....

ofZach:
I've added a PR putting this in the success modal (using the code that @Daandelange wrote) for now: 

https://github.com/openframeworks/projectGenerator/pull/73

I think after 0.9 we can revisit the overall UI and improve....

ofZach:
this has been added, thanks for the suggestion!   please test !

PR #73, opened 2015-10-29
open in IDE

ofZach:
this adds an open in IDE button on success which seems like the simplest way to add this feature (we can work on a better system post 0.9).  It only works on OSX at the moment.

screen shot 2015-10-29 at 9 38 59 am

this PR also fixes a bug with filtering out addons that causes an error when you have the array path wrong (and thus the addon array is null)

### DIFF
--- a/projectGeneratorElectron/app.js
+++ b/projectGeneratorElectron/app.js
@@ -293,6 +293,10 @@ ipc.on('selectAddons', function(arg) {
 //-------------------------------------------
 // allow main to send UI messages
 ipc.on('sendUIMessage', function(arg) {
+
+    // check if it has "success" message: 
+
+
     displayModal(arg);
 });
 
@@ -479,6 +483,11 @@ function setup() {
                 enableAdvancedMode(false);
             }
         });
+
+         $("#IDEButton").on("click", function() {  
+            launchInIDE();
+         });
+
 
          $("#verboseOption").checkbox();
          $("#verboseOption").on("change", function() {
@@ -848,6 +857,13 @@ function displayModal(message) {
 		var shell = require('shell');
 		shell.openExternal( $(this).prop("href") );
     });
+
+    if (message.indexOf("Success!") > -1){
+        $("#IDEButton").show();
+    } else {
+        $("#IDEButton").hide();
+    }
+
     $("#uiModal").modal('show');
 }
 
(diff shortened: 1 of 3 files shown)


ofZach:
@mattfelsen: what is the command to launch the IDE on windows ?  ie, on on osx "open .....xcodeproj"
@arturoc: what is the command to launch the IDE on linux?

mattfelsen:
ooh, shiny! also, great question! i'll see if I can figure it out now. for linux, you can try `xdg-open $DESTINATION_PATH` from @thomasgeissl's  [comment here](https://github.com/openframeworks/projectGenerator/issues/46#issue-109543199)

ofZach:
PS: this PR also puts the actual command line command at the bottom of the IDE message, which could be useful for debugging, etc: 

screen shot 2015-10-29 at 9 50 03 am

mattfelsen:
+1 for that. The parentheses seem like it'd be easy to accidentally copy the closing parenthesis though?

For windows, both of these are working for me:
`start relative\path\to\project\project.sln`
or
`start C:\absolute\path\to\project\project.sln`

ofZach:
ok added windows and linux now, invites testing ! 

@mattfelsen do you think I should drop the parens?  I kind of thought it was weird without context but maybe we can say command used:\n....... instead of parentheses)

ofZach:
PS: for basic testing, if you are not building electron, you can replace out index.html, index.js and app.js in the nightly...

mattfelsen:
Button is working for me, both using `npm start` and running the exe from the dist folder.

On a UX note, I think hitting the button should close the modal, just as the Close button does. It seems weird to close visual studio and see the pg with that modal still up.

And yea, no parens and a newline is what I was thinking :blush:

ofZach:
good idea about closing the modal...  just added that.

will take a look at the parens in a sec. thanks for the feedback!

arturoc:
we don't use codeblocks anymore in linux, the extension is qbs for qtcreator.

ofZach:
oh ok !  let me update the labels...  is this accurate: 

screen shot 2015-10-29 at 10 14 15 am

arturoc:
yes

ofZach:
ok made that change !  thanks !

mattfelsen:
very fond of [this line](https://github.com/openframeworks/projectGenerator/pull/73/files#diff-8cef9e377939ac98ef1630a154e68a07R935): `console.log("sdfsadfasdf?");`

ofZach:
oof thanks -- cleaned it up

ofZach:
going to merge this in so folks can test on the nightlies, etc...  I think it's ok, there might be some weird edge cases like when you build with multiple platforms or make a project that's not your default platform...

Daandelange:
Tested and works great here :)

PR #74, opened 2015-10-29
this is a good icon

ofZach:
this adds a missing .icns file.  On osx this work, but I'm not 100% sure about icons on other platforms: 

screen shot 2015-10-29 at 10 07 02 am


ofZach:
this PR has a funny name, but you get the idea :)

mattfelsen:
I might be opening a whole can of worms here, but... I kind of got attached to that circular electron one with the white border, so I took a quick stab and marrying that with the OF logo. Good/bad? Also I don't know how to turn this into an .icns file but here's a 1024x1024 png

of-electron

ofZach:
ok !  switched it up (I use icns builder)....

now looks like: 

screen shot 2015-10-29 at 10 32 40 am

mattfelsen:
hawt

ofZach:
merging this in now...

Issue #80, opened 2015-11-09
A Javascript error occured in browser process


KavinHan:
**When i click this button:**
![image](https://cloud.githubusercontent.com/assets/3361694/11024644/d3d2387a-86d4-11e5-886a-b46080250815.png)

**I get this error:**
![image](https://cloud.githubusercontent.com/assets/3361694/11024663/141c3890-86d5-11e5-99ba-cda2a1dd05bb.png)

-----
My PC:
OS: windows 10, 64-bit
CPU: i7-4770
oF: 0.9.0_vs_release

ofZach:
thanks! I've been able to recreate this over here...   have just submitted a PR which should help.

ofZach:
I've merged this in -- if you are feeling adventurous -- feel free to change this line in index.js inside the project generator folder (I can't remember exactly where is located on windows but maybe in resources/app): 

https://github.com/openframeworks/projectGenerator/pull/81/files

and let me know if it helps -- otherwise, this change should be in the nightlies shortly (which you can find on the openFrameworks download page...  might take another 24 hours to wind up there)

ofZach:
(ping @Daandelange since this seems like an error we both saw at some points)

KavinHan:
Thank you @ofZach. 

this commit works good.! ^^
https://github.com/openframeworks/projectGenerator/pull/81/files

I think this issues can close~~

ofZach:
thanks!  closing this then...   thanks for the reporting and testing

PR #81, opened 2015-11-09
fix for join not working

ofZach:
this is a potential fix for https://github.com/openframeworks/projectGenerator/issues/80


ofZach:
@Daandelange -- quick ping, this looks right, right?  (I remember we had to use temp path objects elsewhere for similar errors)....

ofZach:
I'm going to merge this for now since it seems to do the right thing to me over here and we can test the nightlies once it's in...

Daandelange:
(yes, looks OK to me)

PR #84, opened 2015-11-09
Fix select addons list height problem

KavinHan:
Problem:
**If selected addons is empty then looks is ok. 
you can see ofxAffineTransformation addon at top**

**But if has some selected addons then have a problem. 
The ofxAffineTransformation was display at behind of top header bar. 
So i change the select height to `max-height: 16em !important;`**

---

I'm sorry about my terrible English...=,=
Can u understand this problem?...May be u have a better solution. I just try to fix it.^^


danoli3:
thanks this was fixed

PR #86, opened 2015-11-09
Added command line target to xcode project

HalfdanJ:
I've created a new target in the command line xcode project that compiles directly to a command line application, instead to an .app


arturoc:
we should be able to reproduce this project file automatically. right now, the PG is reading the config.make file for an after rule and adding that to the osx project, also we use the nofmod template to create the project with the right settings. there could be a template that has this changes so the PG can create the project with this structure from that template

HalfdanJ:
Okay, makes sense to try and implement this pr in the template instead then.

PR #90, opened 2015-11-11
Add addon dependencies

### LINKED ISSUE #30: ADDON_DEPENDENCIES from addon_config.mk is not processed
Thus, when an addon only lists it's primary addon (e.g. `ofxHTTP`), the dependent addons listed in the addons_config.mk file are not tracked down.  Ultimately this will probably have to happen recursively making sure that the whole tree is traversed.  

The makefile system respects the `ADDON_DEPENDENCIES` var and adds the required addons, so this fix will bring those two systems in line with one another.

I'm happy to take a look and implement this, but as there is active PG development at the moment, I don't want to interrupt until things calm just a bit.

HalfdanJ:
Recursively add addon dependencies Fixes #30

### DIFF
--- a/ofxProjectGenerator/src/projects/baseProject.cpp
+++ b/ofxProjectGenerator/src/projects/baseProject.cpp
@@ -260,6 +260,10 @@ void baseProject::addAddon(ofAddon & addon){
     for(int i=0;i<(int)addons.size();i++){
 		if(addons[i].name==addon.name) return;
 	}
+    
+    for(int i=0;i<addon.dependencies.size();i++){
+        addAddon(addon.dependencies[i]);
+    }
 
 	addons.push_back(addon);
 
--- a/ofxProjectGenerator/src/projects/visualStudioProject.cpp
+++ b/ofxProjectGenerator/src/projects/visualStudioProject.cpp
@@ -341,6 +341,10 @@ void visualStudioProject::addAddon(ofAddon & addon){
 		if(addons[i].name==addon.name) return;
 	}
 
+    for(int i=0;i<addon.dependencies.size();i++){
+        baseProject::addAddon(addon.dependencies[i]);
+    }
+
 	addons.push_back(addon);
 
     for(int i=0;i<(int)addon.includePaths.size();i++){
--- a/ofxProjectGenerator/src/projects/xcodeProject.cpp
+++ b/ofxProjectGenerator/src/projects/xcodeProject.cpp
@@ -1165,6 +1165,10 @@ void xcodeProject::addAddon(ofAddon & addon){
 		if(addons[i].name==addon.name) return;
 	}
 
+    for(int i=0;i<addon.dependencies.size();i++){
+        baseProject::addAddon(addon.dependencies[i]);
+    }
+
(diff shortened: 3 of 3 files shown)


ofZach:
thanks this is great !  I'd like to check how the PG handles multiple addons, like if ofxCv include ofxOpenCv but ofxOpenCv is already in the addons.make file, does it get included twice?  I am wondering if need some smarts or not....

HalfdanJ:
It needs a check yes. And i just realized that the make system has the same problem (just tried with ofxCV depending on ofxOpenCv..)

arturoc:
instead of adding a check it's easier and cleaner to use a set for the collection of addons that way every addon no matter how many times is added will be unique. the return for the method that gets the collection of addons can still be a vector that gets filled from the set

HalfdanJ:
I just looked into this, the check is already there: https://github.com/openframeworks/projectGenerator/blob/master/ofxProjectGenerator/src/projects/baseProject.cpp#L261 

Should i change it to the set solution @arturoc mentions, or should we keep it as it is?

bakercp:
Just wanted to ping this issue -- any way I can help move it though?

bakercp:
Ping.  Would love to see this merged ...

arturoc:
i think this might need some work still to avoid duplicated addons. i think adding the addons in a set as a first step to then add them should solve any potential problem. i guess we could have a public method that is 

```
addAddons(vector<std::string>)
```

so all addons have to be added at once and then that method takes care of parsing the dependencies and finally add the individual addons one by one calling the now private `addAddon`

the command line tool already has a vector<string> for the addons so the change would be trivial

bakercp:
This is becoming more important now for automated unit testing etc.

@thomasgeissl has done some work on this interesting project that seems related -- https://github.com/thomasgeissl/ofPackageManager

While I'm not super interested in creating a npm / bundler-style dependency manager it would be really nice to be able to handle more slightly more complex dependency lists in the addon_config.make file and the addons.make ... 

At the very least it seems that we should be able specify the github user / repo e.g. `bakercp/ofxHTTP` and branch / tag.  Thoughts on the best way to do this? (related https://github.com/openframeworks/projectGenerator/issues/30)

bakercp:
Here's a nice well-articulated syntax that I'd love to support in `addon_config.mk` and `addons.make` files:

https://docs.npmjs.com/files/package.json#git-urls-as-dependencies
https://docs.npmjs.com/files/package.json#github-urls

And we kind of already support this with the `local_addons` syntax, etc.
https://docs.npmjs.com/files/package.json#local-paths

ofTheo:
Just ran into this - would love to get this / #30 fixed. 
I can see the possible recursion issues people mention. 

Probably just checking the addons vector before adding and ignore duplicates will solve this. 
I can submit a PR to your branch @HalfdanJ .

ofTheo:
Actually this looks fine as is. It avoid duplicates at the start of that function already. 
Going to merge this - there might be better ways in the future but this is a good start.

Autr:
Is this in the latest 0.11 release? I doesn't happen with the bundled PG...

PR #93, opened 2015-11-16
Fix paths for local addons when parsing addon_config.mk

mattfelsen:
When parsing paths from addon_config.mk for things like ADDON_INCLUDES, ADDON_SOURCES, etc. in local addons, the path was still relative to OF instead of the addon. For example, this is what I was getting with this [addon_config.mk file](https://github.com/labatrockwell/ofxLibwebsockets/blob/79e4afce94137b13166825d9f66901a015a93dfe/addon_config.mk#L16-L22) from [ofxLibwebsockets/0.9.0](https://github.com/labatrockwell/ofxLibwebsockets/tree/0.9.0)

config-bad

You can see the paths for the other addons, which are picked up from the filesystem and not from addon_config.mk, are properly resolved. And with this fix:

config-good

### DIFF
--- a/ofxProjectGenerator/src/addons/ofAddon.cpp
+++ b/ofxProjectGenerator/src/addons/ofAddon.cpp
@@ -194,7 +194,10 @@ void ofAddon::parseVariableValue(string variable, string value, bool addToValue,
 		return;
 	}
 
-	string addonRelPath = ofFilePath::addTrailingSlash(pathToOF) + "addons/" + name;
+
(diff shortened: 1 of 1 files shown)


ofZach:
thanks seems like a good fix, merging....

mattfelsen:
Well that was fast. Thanks!

ofZach:
:) benefit of such a small PR!  well documented problem and easy to understand solution....  tx

PR #94, opened 2015-11-24
Update main.cpp

javl:
* More verbose output (whole sentences, capitals, etc.)
* Added --force argument (was already in code, but not accepted as argument)
* Added --available-platforms argument that returns a (hard-coded) list of available build platforms)
* Updated examples
* Don't throw error when called with only --help argument

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -4,19 +4,21 @@
 #include <string>
 #include <vector>
 #include "optionparser.h"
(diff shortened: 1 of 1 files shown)


javl:
The Travis CI build failed, caused by some SSL error, which doesn't seem to be related to my pull request. Is it possible to somehow restart the check?
Also: bump

arturoc:
can you pull from master and push again? there was a problem with the travis builds but it should be fixed now

PR #97, opened 2015-12-02
add addonCFlags to OTHER_CFLAGS in Xcode, fixes #96

### LINKED ISSUE #96: OTHER_CFLAGS not set when generating Xcode projects
OTHER_CFLAGS are not being set when generating Xcode projects. This means ofxPd Xcode projects will not build in 0.9.0 ;(

I had this fixed with https://github.com/ofZach/projectGeneratorSimple/pull/62, but it seems not everything made it into this repo. I believe the fix is to change `OTHER_CPLUSPLUSFLAGS` to `OTHER_CFLAGS` as in my original PR on xcodeProject.cpp line 1053: https://github.com/openframeworks/projectGenerator/blob/master/ofxProjectGenerator/src/projects/xcodeProject.cpp#L1053

danomatika:
Fixes #96


arturoc:
thanks!

Issue #82, opened 2015-11-09
time to delete the simple and advanced tools


ofZach:
feels like the command line tool and gui are working well-ish now...  we should remove the older legacy tools

arturoc:
yeah let's do it :)

danoli3:
Delete it all confusing! However make steps to make this new sexy generator.

mattfelsen:
Fixed via #106, this can be closed

PR #103, opened 2015-12-15
projectGeneratorSimple - Fixes missing iOS OS X Target

danoli3:
Currently to build for iOS (OS X App) you had to either build from command line or manually enter the pre-processor (not easily found / known).

This adds targets for iOS / OS X (of course all are still OS X Apps).

screen shot 2015-12-16 at 4 46 59 am


ofZach:
thanks!  we are planning to remove pgSimple from the repo as per https://github.com/openframeworks/projectGenerator/issues/82
(since we are using the command line tool and the electron based gui now)

danoli3:
OMGGG I just downloaded the nightly and the project generator is so sexy... how do I build this?!!!

ofZach:
It's in the 0.9 download :) but yes more documentation would be good, let me know what will help.

Issue #104, opened 2015-12-15
Steps to Build Electron?


danoli3:
So to Build the Electron project are the following steps correct:

```
npm install electron-packager --save-dev
npm install electron-packager -g
```

Then the following:

``` 
npm install

nam start

npm run build

```

Then how do you specify what target?

mattfelsen:
Hey @danoli3. I think `electron-packager` gets installed when you run `npm install`. I don't remember running in manually myself.

Yes on `npm install` to install all the modules and `npm start` to get it running.

`npm run build` will build for all platforms. I think that this is supposed to work, but I believe I got an error when it was trying to package for linux..

platform-specifc variants are...
- `npm run build:osx`
- `npm run build:linux32`
- `npm run build:linux64`
- `npm run build:vs`

You can see them if you take a look at the `scripts` section inside `package.json`

arturoc:
you can also find the very latest build incluing the binary for the command line tool in the correct place here

http://ci.openframeworks.cc/projectGenerator/

the ci servers build the PG whenever there's any change and put it there for the nightlies packaging tool.

mattfelsen:
I think this can get closed now that #105 is merged

Issue #83, opened 2015-11-09
merge with openFrameworks core


ofZach:
@ofTheo reported that it's still not super clear with the project generator how to make examples for OF when you pull from github, I feel like moving this submodule back into the core of OF would help make things more clear and easier to track the progress of this project.     it's also been tricky to have PRs across both the main OF repo and this repo when things like templates and the addon change, as well as sometimes the submodule link is out of date.  

I'm in favor of bringing this back into the core of OF and making it a bit easier to understand and use the command line tool for updating the OF examples.

HalfdanJ:
Something that would really help is a overview readme of this repo. I'm still confused on the different modules in the root of this repo.

HalfdanJ:
Also based on personal experience, i think this is mostly about having a good readme in openframeworks root that describes properly how to set up the development enviroment for openframeworks, including what commands to get going with generating the examples.

ofTheo:
would love this! 
It took @nickhardeman and I a couple of hours, including installing different versions of Git command line tools to get the submodule working.  My feeling was if it was this hard for us to do it, how many other people would have gotten frustrated and given up. 

I think a project generator label in the OF repo is all thats needed for filtering the PG issues. 
Also I think now that the big push has been made on PG - it won't really need a separate repo. 

Also at this point could we also remove projectGeneratorLegacy and Simple?

ofZach:
yep on it -- https://github.com/openframeworks/projectGenerator/issues/82

I wonder if we could have a specific readme for developers and a readme for end users....  The developer readme would not be in the packaged release, but would give very clear info about how to setup the dev environment, etc...  the end user readme could be more friendly like where to ask questions, ask for help, etc.    It could also be more exhaustive and link back to the website docs which would be more up to date.

HalfdanJ:
I think that would be great. Would be great if the readme on github was more focused on oF development then usage, and just pointing "end users" in the right direction (openframeworks.cc)

elliotwoods:
i've hit into this a couple of times
finally realised the best way for me is to copy it out of the downloadable oF zip 
(i.e. i want a working version of oF from git, but don't want to deal with node.js)
kinda feels dirty, so just putting it out there so others don't feel bad for doing the same :)

mattfelsen:
@elliotwoods jaw saw Arturo mention this over on #104...you can get a pg binary built from github from http://ci.openframeworks.cc/projectGenerator/ without having to download the whole OF release/nightly :+1:

[...]

PR #105, opened 2015-12-20
Update readme for electron app

mattfelsen:
First pass at adding some documentation on how to get the electron fronted up & running. Some of the docs will probably soon be outdated if folders are renamed (which I do in my forthcoming PR) and there are some steps which should get automated or have scripts for, but I believe this is what's necessary as-is, so we gotta start somewhere :smile:


mattfelsen:
Helpful for issues like https://github.com/openframeworks/projectGenerator/issues/104 and https://github.com/openframeworks/projectGenerator/issues/83#issuecomment-155059903

mattfelsen:
Pinging @danoli3 too who dug into this recently, would be good to know if this is a helpful format

danoli3:
Awesome thanks! Makes a lot more sense :D!

ofZach:
this looks great - thanks! merging!

Daandelange:
Thanks, nice start! 
Maybe we could also put a link to [npm](https://nodejs.org/en/download/) which is not installed by default.

mattfelsen:
Oh good point, I always forget that about node/npm. I'll add a note and
suggest a couple ways to install:

- easiest: from nodejs.org
- homebrew
- homebrew + npm

Is anyone aware of any node version dependencies? I've used & built the
electron app with 0.10 and 5.3 so I think we're in the clear, but just
checking

mattfelsen:
Whopps, that's homebrew + *nvm*. Corrected

ofZach:
one thing I'd like to mention in the readme (I am making an issue here just to remind myself to add a note) is the use of "npm-shrinkwrap.json" ie: 

https://github.com/openframeworks/projectGenerator/blob/master/projectGeneratorElectron/npm-shrinkwrap.json

to fix a bug that was building packages on osx incorrectly so they were failing on 10.8.   I think it's useful to mention this in case anyone starts manipulating the modules / etc as I'm not sure if shrinkwrap needs to be regenerated, etc. 

(also this bug might be fixed by now so if we can remove this, it make the whole thing less fragile) 

I need to do a bit of research to remember the problem and solution but I'll add this shortly.

mattfelsen:
@ofZach I played around with shrinkwrap for a bit but could not get it to work! `npm install` works and the app starts fine, but then trying `npm shrinkwrap` from my working setup kept giving errors, mostly around a version mismatch between electron-debug 0.1.1 being wanted but electron-debug 0.2.1 being installed? Also it looks like [devDependencies are ignored for shrinkwrapping](https://docs.npmjs.com/cli/shrinkwrap#other-notes), so those might need to be moved to dependencies (I tried this, and shrinkwrap is still complaining).

So I can't really test this, and I'm not sure how to test/reproduce the issues you saw on 10.8, so I'll leave it to you to figure out :smile: Here's the bit I wrote for the readme which I'm leaving out for now, but feel free to re-use later if you like:

### This package is shrinkwrapped!

We use `npm shrinkwrap` to specify versions not only of the node modules we use directly, but their dependencies as well. If you are working on the electron app and need to add npm modules, be sure to update the shrinkwrap after and include `npm-shrinkwrap.json` in your pull request along with `package.json` and any code updates.

```
# the usual
npm install --save new-module

# do some coding, make sure everything works

# shrinkwrap it!
npm shrinkwrap

# add npm-shrinkwrap.json when you commit!
```

ofZach:
thanks it's gonna take me a few days to get my gears cranking again.  also it looks like this issue: 

https://github.com/maxogden/electron-packager/issues/111

has been fixed upstream

https://github.com/maxogden/extract-zip/issues/6

so we might not even need shrinkwrap....   It was basically being used to point to a specific fix upstream.   gonna check....

mattfelsen:
Gotcha. All good, I'm on a pg roll today :sunglasses:

Issue #118, opened 2016-03-02
problem with iOS project creation


arturoc:
the logs for the nightly builds are reporting the following problem when generating iOS projects:

```
[ error ] ofFile: copyTo(): source file /home/ofadmin/openFrameworks/scripts/dev/openFrameworks_pkg_creation/scripts/templates/ios/bin/data/Default-568h@2x~iphone.png does not exist
[ error ] ofFile: copyTo(): source file /home/ofadmin/openFrameworks/scripts/dev/openFrameworks_pkg_creation/scripts/templates/ios/bin/data/Default.png does not exist
[ error ] ofFile: copyTo(): source file /home/ofadmin/openFrameworks/scripts/dev/openFrameworks_pkg_creation/scripts/templates/ios/bin/data/Default@2x.png does not exist
[ error ] ofFile: copyTo(): source file /home/ofadmin/openFrameworks/scripts/dev/openFrameworks_pkg_creation/scripts/templates/ios/bin/data/Icon.png does not exist
[ error ] ofFile: copyTo(): source file /home/ofadmin/openFrameworks/scripts/dev/openFrameworks_pkg_creation/scripts/templates/ios/bin/data/Icon@2x.png does not exist
```

Not sure where the error is coming from or if it's critical though. Seems strange that it's trying to copy a file from the templates that wouldn't exist so it might be some error in the path solving when copying or something similar

ofZach:
hmm. pinging @danoli3 - I think it has to do with this: 

https://github.com/openframeworks/openFrameworks/commit/7af3e9bae700f9308ba5d8a5aa75a5a3dc133d00

and needing to maybe update this: 

https://github.com/openframeworks/projectGenerator/blame/master/ofxProjectGenerator/src/projects/xcodeProject.cpp#L344-L359

arturoc:
yeah looks like, that last bit in the PG code should probably copy all the contents of the folder using ofDirectory instead of hardcoding the exact files

danoli3:
Ah yes! I was meant to update this when we merged the new splash / icons! I'll PR to Project generator now. I have it already done.

PR #120, opened 2016-03-03
Fix for new mediaAssets iOS

danoli3:
Fixes https://github.com/openframeworks/projectGenerator/issues/118

Copy's directory of the new iOS mediaAssets folder if it doesn't exist.

### DIFF
--- a/ofxProjectGenerator/src/projects/xcodeProject.cpp
+++ b/ofxProjectGenerator/src/projects/xcodeProject.cpp
@@ -341,20 +341,10 @@ bool xcodeProject::createProjectFile(){
 				dataDirectory.create(false);
 			}
 		}
-        if(!ofFile::doesFileExist(projectDir + "/bin/data/Default-568h@2x~iphone.png", true)) {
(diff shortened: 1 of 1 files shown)


arturoc:
This is failing to compile. Also, i think we probably should copy the files recursively instead of the full path. i'm not 100% sure but copying the full folder might overwrite whatever was there before which would delete any files you might already have there when updating a project

danoli3:
Yeah fixed that, minor function name mess-up.

It does 2 checks:
``` 
if (!mediaAssetsProjectDirectory.exists()){
```
If the mediaAssetsProjectDirectory exists it won't even copy it.

The second is embedded in the command:
> ...copyTo(mediaAssetsProjectDirectory.getAbsolutePath(), false, **false**);
Which is overwrite=false 

---

Is there an alternative recursive overwrite only files that don't exist command with ofDirectory?

arturoc:
no, i'm checking and ofDirectory::copyTo just deletes the destination folder if it exists and you pass overwrite. 

but thinking about it, i would say it's ok like that, it's the same as we've talked about the src folder, if it's already there it's better to not touch it. someone might have changed those files with their own icons or whatever and that would overwrite them.

danoli3:
Yeah, good to merge then?

arturoc:
yeah, thanks!

PR #124, opened 2016-03-18
print correct running times

brendan-w:
The commandLine project generator was reporting large, negative elapsed times:

```
[notice ] setting OF path to: /home/brendan/of_v0.9.3_linux64_release/
[notice ] from -o option
[notice ] target platform is: linux64
[notice ] updating project /home/brendan/of_v0.9.3_linux64_release/apps/myApps/test
[notice ] saving addons.make
[notice ] project updated! 

1 project updated [notice ] in -8528.8 seconds
```

This was due to the fact that OF's start time values are being set in [this call to `initutils()`](https://github.com/openframeworks/projectGenerator/blob/master/commandLine/src/main.cpp#L439)  _after_ the [start time had been recorded](https://github.com/openframeworks/projectGenerator/blob/master/commandLine/src/main.cpp#L436). Since these calls were so close together, I thought it simplest to use only one `ofGetElapsedTimef()`, and let `initutils()` mark the start time for us.

The project generator now outputs:

```
[notice ] setting OF path to: /home/brendan/of_v0.9.3_linux64_release/
[notice ] from -o option
[notice ] target platform is: linux64
[notice ] updating project /home/brendan/of_v0.9.3_linux64_release/apps/myApps/test
[notice ] saving addons.make
[notice ] project updated! 


1 project updated [notice ] in 0.218913 seconds
```

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -52,7 +52,6 @@ enum pgMode {
 };
 
 
-float               startTime;
 int                 nProjectsUpdated;
 int                 nProjectsCreated;
 
@@ -352,7 +351,6 @@ int main(int argc, char* argv[]){
     bHelpRequested = false;
     bListTemplates = false;
     targets.push_back(ofGetTargetPlatform());
-    startTime = 0;
     nProjectsUpdated = 0;
     nProjectsCreated = 0;
     string projectName = "";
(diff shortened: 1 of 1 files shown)


arturoc:
Closing this since another simpler solution was merged already

PR #142, opened 2016-12-01
VS: Fix issue with .dlls in addons

tgfrerer:
These changes affect only projects generated for Visual Studio/Windows.

Proposes some changes that make it possible to use addons, and local addons which provide `.dlls` inside their libs. 

Dlls will need to be specified in `addon_config.mk` using the existing entry `ADDON_DLLS_TO_COPY`. Any `.dll` specified this way will be copied into a generated app's `bin` directory, so it can be found during the linker stage.

* also fixes an issue where the project generator would attempt to link against `.dll` files instead of just `.lib` files: when parsing addon libs, the updated project generator will skip any library files which are not `.lib` files for vs.

Otherwise, our current version of ofxAssimpModellLoader for example, will fail to link.

### DIFF
--- a/ofxProjectGenerator/src/projects/visualStudioProject.cpp
+++ b/ofxProjectGenerator/src/projects/visualStudioProject.cpp
@@ -348,16 +348,6 @@ void visualStudioProject::addAddon(ofAddon & addon){
         addInclude(addon.includePaths[i]);
     }
 
-    // divide libs into debug and release libs
-    // the most reliable would be to have seperate
-    // folders for debug and release libs
-    // i'm gonna start with a ghetto approach of just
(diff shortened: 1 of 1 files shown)


arturoc:
probably by default we should copy any dll present in the libs folder then overwrite that with the ADDON_DLLS_TO_COPY variable if using = or add to that if using += as it happens with other variables. first parse from the file system then add to that or overwrite using the addon config.

tgfrerer:
Yes, that makes perfect sense - I'll update the or when I get to it next...

arturoc:
just had a moment to check this. i've added a rule to copy all the dlls present in the filesystem too. thanks!

PR #145, opened 2017-01-28
Add simple readme

kr15h:
There were problems compiling this and we figured it out during the workshop and I just wanted to post it here instead of somewhere deep in the forums.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,9 @@
+# Project Generator
+
+To compile on OS X, one has to cd into the projectGenerator directory and execute the following.
+
+```
+./scripts/osx/buildPG.sh
+```
+
+Should work.


arturoc:
Those scripts are really meant to be used on the servers that build the release although they should work on any computer really, the usual ways of building a project, using makefiles or the xcode or qtcreator projects should work fine for the project generator too.

which problems where you having?

kkshmz:
Is this not going to be added or ?

kr15h:
Will iterate on this.

haschdl:
@arturoc @kr15h  is the proposed README containing outdated/incorrect info? maybe just reject the PR, it's two years old now :)

PR #152, opened 2017-03-27
Linux CI deploy script fix

HalfdanJ:
....


arturoc:
feel free to commit directly to master to test this if it keeps failing, or you can also disable the check that only tries to upload if not on a PR and in master until it works and then reenable again before merging to master

PR #154, opened 2017-04-04
Update readme.md

bangnoise:
Fix examples so they actually work


arturoc:
can you change it to just remove the space? i think that's more standard syntax and would be valid to if we ever change the arguments parser. thanks!

bangnoise:
Okay, there you go

arturoc:
thanks!

PR #161, opened 2017-06-14
Fixes #111 (XCode template Default Western Encoding -> UFT8)

### LINKED ISSUE #111: pg is not utf-8 friendly on OSX
I've heard reports that utf-8 files are getting encoded wrong when added to the project.  Also, the src code for the template is encoded as "western".

I'm pretty sure the fix is two part: 

a) this encoding choice

https://github.com/openframeworks/projectGenerator/blob/master/ofxProjectGenerator/src/projects/xcodeProject.cpp#L56-L57

should be 4 (utf-8) and not 30 so incoming src files are treated as utf-8.  

b) src files for the template need to be changed as well to utf-8

it would be good to have a tests cases to study, such as an addon that has japanese comments, etc.

bakercp:
Fixes #111 
Fixes https://github.com/openframeworks/openFrameworks/issues/4728

https://github.com/openframeworks/openFrameworks/pull/5664

### DIFF
--- a/ofxProjectGenerator/src/projects/xcodeProject.cpp
+++ b/ofxProjectGenerator/src/projects/xcodeProject.cpp
@@ -54,7 +54,7 @@ STRINGIFY(
             <key>explicitFileType</key>
             <string>FILETYPE</string>
             <key>fileEncoding</key>
-            <string>30</string>
+            <string>4</string>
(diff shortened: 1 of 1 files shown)


arturoc:
thanks!

PR #162, opened 2017-06-16
Removing using namespace std

Paul92:
Removing ```using namespace std``` from the core library (see [PR](https://github.com/openframeworks/openFrameworks/pull/5671)) broke the project generator. I fixed the errors caused by that, now it compiles.

Needs more testing. Also, there are a few more occurences of ```using namespace std``` in the project generator itself - i'm looking forward to remove them too if there is interest in that.


arturoc:
thanks!

Issue #163, opened 2017-08-07
Xcode Projects don't throw error when method is supposed to return but it doesn't.


armadillu:
So a method like:

```
bool doSomething(){

}
```

Doesn't lead to a compile error on Xcode with the current setup. This is very easily overlooked, especially for beginners; I think I'd much rather have an easy-to-fix compiler error than spending hours wondering why my code behaves randomly. Adding a couple compiler flags resolves the situation. 

```
-Wreturn-type -Werror=return-type
```

I'll prepare a PR if that's cool.

https://github.com/armadillu/projectGenerator/commit/6a7364a1c46bd775f924cc1724279e5b4256a8f9

arturoc:
yes this is the default in c++ but it's terribly error prone, i recently enabled a similar flag on qtcreator and the makefiles. there's a couple of cases where you get false positives because switch/cases mostly but i think it's worth enabling it.  if you can send a PR?

armadillu:
Sent - https://github.com/openframeworks/projectGenerator/pull/164

PR #164, opened 2017-08-08
xcode prj defaults to error when method that's supposed to return doe…

armadillu:
…sn't (instead of warning)

As discussed in #163 - add compiler flags to Xcode projects created with project generator so that methods that miss a return value trigger compiler errors instead of warnings.


armadillu:
Hey @arturo does this look good?

arturoc:
yeah sorry, i was waiting for the tests to pass that day and then completely forgot. thanks!

PR #166, opened 2017-08-30
Feature/support xcode defines

armadillu:
This adds support for 
```
	ADDON_DEFINES = MY_DEFINE1 MY_DEFINE2
```

on Xcode projects.

Note that it doesn't handle

```
	ADDON_DEFINES = MY_DEFINE=VALUE
```
properly yet; I've looked into that but it would require a much more invasive change to handle that as the '=' token is expected only for defining variables.

### DIFF
--- a/ofxProjectGenerator/src/projects/xcodeProject.cpp
+++ b/ofxProjectGenerator/src/projects/xcodeProject.cpp
@@ -171,6 +171,15 @@ STRINGIFY(
 
 );
 
+//-----------------------------------------------------------------
+const char Defines[] =
+STRINGIFY(
+
+          <key>GCC_PREPROCESSOR_DEFINITIONS</key>
+          <array></array>
+
+);
+
 const char workspace[] =
 STRINGIFY(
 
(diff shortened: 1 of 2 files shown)


armadillu:
It seems the removal of std:: from global space is breaking this; I saw there's another PR supposedly fixing this.

arturoc:
can you rebase this to make it mergeable again?

armadillu:
yep coming up - ill rebase the other PR over the weekend

arturoc:
thanks!

eduardfrigola:
Hi! I'm on macOS 10.13.6, Xcode version 10.1, last project generator compiled version from of0.10.1 release.
When defining ADDON_DEFINES = OFXOCEANODE_USE_OSC (in my case) breaks generation of the project. Other test is that when defining only ADDON_DEFINES = (without specifying flags) it generates the project successfully.

Log with PG app:
```
Command failed: /bin/sh -c 
"/Users/eduardfrigola/Documents/Dev/openFrameworks/macOS_git/projectGenerator/projectGenerator.app/Contents/Resources/app/app/projectGenerator" -v -o"/Users/eduardfrigola/Documents/Dev/openFrameworks/macOS_git" -a"ofxDatGui_PM,ofxOceanode,ofxOceanode_Osc,ofxOsc" -p"osx" "/Users/eduardfrigola/Documents/Dev/openFrameworks/macOS_git/apps/Playmodes/oceanodeOscAutoTest"
```

Log from command line PG:
```
[verbose] in fromFS, trying src /Users/eduardfrigola/Documents/Dev/openFrameworks/macOS_git/addons/ofxOceanode_Osc/src
[verbose] ofAddon: libs after exclusions 0
[notice ] adding addon ofxOceanode_Osc
Segmentation fault: 11
```

The addon that has the errors is an addon_config only addon, intended to use for defining macros for another addon.

Can you point me to the right place to be able to try and fix this?

Thanks!
Eduard

armadillu:
Hi @eduardfrigola, I've just tested the PG included in the 10.1 OSX release to generate  project with [ofxFontStash](https://github.com/armadillu/ofxFontStash)  and it worked ok. 

ofxFontStash uses the [ADDON_DEFINES](https://github.com/armadillu/ofxFontStash/blob/master/addon_config.mk) clause (that's why I made the PR to begin with) so maybe the problem is in ofxOceanode_osc? I can't the addon in your github profile btw, do you have it hosted somewhere?

eduardfrigola:
Hi @armadillu thanks for your feedback, I've published it to github: https://github.com/PlaymodesStudio/ofxOceanode_Osc
If I use it defining CFLAGS it works (adds C flags to xcode project)

Thanks!

armadillu:
@eduardfrigola That's only the .mk file, can you share the whole addon? Just for a quick test I dropped your addon_config.mk file into ofxFontStash and it worked fine, so I suspect it the issue may not be the  ADDON_DEFINES

eduardfrigola:
It is the hole addon... I think that is where the error come from...
If I make the same as you and put it in another addon it works.
The strange thing is that defining other variables, like CFLAGS work with this single file addon...

Maybe the problem is because is an empty addon and not in your implementation of ADDON_DEFINES itself...

armadillu:
Whait what's the point of the addon then? Just to create the FLAG? Oh yeah I think PG assumes some files will be there. It's probably crashing before it tries to add the Defines...

eduardfrigola:
But the strange thing is that if I add flags it does not crash and it adds the flags correctly...

eduardfrigola:
Hi Oriol!
I checked with PG from of0.11 release and I do not have the segmentation fault now, so problem solved.

Sorry I never answered your question about why I want to use that.
Something similar as memo discuses here: https://forum.openframeworks.cc/t/conditional-compiles-for-addons/1209

I have some functionality in my addon: [ofxOceanode](https://github.com/playmodesstudio/ofxOceanode/), that I want to be optional (for example osc or midi functionality). I did that with macros, for the addon not looking for osc or midi files when compiling. With a only configuration file i can easily activate the functionaly inside my addon and also add the necessary addons with ADDON_DEPENDENCIES.
So that is why an addon with just .mk file makes sense for me, knowing that maybe is not the best way to deal with this.

Thanks for your time in this!

PR #168, opened 2017-08-31
Bugfix/visual studio defines

armadillu:
support for defines persisting across multiple addons - the old code would only keep the latest define.


armadillu:
same here, the other PR with the std namespace will probably fix this.

PR #178, opened 2018-05-20
fix: call ofGetElapsedTimef after initutils()

2bbb:


2bbb:
this is other solution of #124

PR #184, opened 2018-08-07
fix: enable some compiler flag including `=`

2bbb:
example:

```
ADDON_CPPFLAGS = "-Werror=return-type"
```

now, this flag will get wrong result like `-Werror`.

this PR fixes those problem.


arturoc:
thanks!

PR #190, opened 2018-09-25
Support ADDON_DATA in addon_config.mk

hiroMTB:
## original PR
https://github.com/openframeworks/projectGenerator/pull/92 by @mattfelsen in 2015-16.

## additional commit
I checked original PR still works nicely but only able to copy single file.
I added folder copy feature so that we can avoid listing up all files.
Tested on OSX and Windows, both of commandLine and frontend app works okay.

## Known problem
If we have same data folder/file in different addon, PG will copy both to /myProject/bin/data but overwrite. For example, considering we have following two addons. 
- ofxCoolShaders
  - /src
  - /data/img
- ofxCoolPhotos
  - /src
  - /data/img

And generate a project which copies both of /data/img. Then /myProject/bin/data/img folder will be overwritten by last addon specified. This might be solved by adding new "merge" feature to ofDirectory class in order to avoid overwrite folder but merge contents.

### DIFF
--- a/ofxProjectGenerator/src/addons/ofAddon.cpp
+++ b/ofxProjectGenerator/src/addons/ofAddon.cpp
@@ -310,7 +310,7 @@ void ofAddon::parseVariableValue(string variable, string value, bool addToValue,
 	}
 
 	if(variable == "ADDON_DATA"){
-		addReplaceStringVector(data,value,addonRelPath,addToValue);
+		addReplaceStringVector(data,value,"",addToValue);
 	}
 
 	if(variable == "ADDON_LIBS_EXCLUDE"){
--- a/ofxProjectGenerator/src/projects/baseProject.cpp
+++ b/ofxProjectGenerator/src/projects/baseProject.cpp
@@ -254,7 +254,23 @@ void baseProject::addAddon(std::string addonName){
         auto standardPath = ofFilePath::join(ofFilePath::join(getOFRoot(), "addons"), addonName);
         addon.fromFS(standardPath, target);
     }
+
     addAddon(addon);
+
+    // Process values from ADDON_DATA
+    if (addon.data.size()) {
+        string dest = ofFilePath::join(projectDir, "bin/data/");
+
+        for (auto& filename : addon.data) {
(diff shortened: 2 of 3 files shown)


arturoc:
thanks!

hiroMTB:
Please notice this can cause problem when user "update" existing project.

It will overwrite his/her own data resource if the folder name is the same.

arturoc:
yeah i saw. i guess most people add addons at the beginning before having any data? also data from addons would usually be named in a way that it's obvious won't clash easily with other files i can imagine

in any case perhaps we could add a check to see if those files exist and fail it they do?

hiroMTB:
I often start a new project and update several times whenever I noticed I need an additional addon. (or I decide I don't need addon already imported)

> also data from addons would usually be named in a way that it's obvious won't clash easily with other files I can imagine

I expect we can find /data/shader in several addons, also /data/png etc?

> in any case perhaps we could add a check to see if those files exist and fail if they do?

We can check existing data folder and cancel copy. But then addon does not work properly because of missing files. I think the best solution is implementing the "merge" feature on ofDirectory. Or we ask addon makers to use a unique folder name. (this sounds a bad idea)

hiroMTB:
Or we can disable folder copy and only allow single file copy (like an original PR).
This case we should list up all files in addon.config with ADDON_DATA.

arturoc:
when i say fail i mean it should show an error so you know that the addon didn't install properly. and yes i think addons with data should have unique folder names like ofxSomething_shaders instead of shaders

hiroMTB:
I see, indeed just skipping is a bad idea.
So, we have four solutions so far.

- A. Only copy folder if the same folder does not exist ( and return an error, fail)
- B. Always copy folder wich has a unique name. (ofxSomething_shaders, overwrite)
- C. Always copy a single file (overwrite)
- D. Always copy with merge feature (overwrite)

I start thinking B is the best.
It makes easy to notice which data come from which addon.
If there is old addon with /data/shader included, probably it does not have ADDON_DATA feature in addon.config. So it won't overwrite user data.

PR #196, opened 2018-11-30
Add props support (for VS)

gumilastik:
I have been waiting for this feature for a long time


arturoc:
can you explain what does this do?

gumilastik:
recursively scans the addon directory, find * .props files and adds them to the project file. many addons use props files for VS and they need to be manually imported each time after generating a project

we can also add props support to addon_config. in this commit, all props files will be imported.

PR #201, opened 2019-03-22
Search Addons for partial match & update semantic.min.js

narikei:
Modify addons search that can only search forward.
Supports partial match search.

but, I needed to update Semantic UI.
As far as I know, this is working properly.
I think the generator will be the best.

### can not found
Screen Shot 2019-03-22 at 20 02 47

### partial match ok
Screen Shot 2019-03-22 at 20 02 23

## referenced
https://semantic-ui.com/introduction/new.html#/twotwo
https://semantic-ui.com/modules/dropdown.html#additional-settings


arturoc:
Nice! thanks!

PR #202, opened 2019-05-17
add vscode support 🚜under development🔨

hiroMTB:
Not finished yet, do not merge, please🙏

🚦 STATUS
- [x] generate project
- [x] update project
- [x] test with addon
- [x] test recursive feature
- [x] test on OSX
- [ ] test on Windows
- [x] test on Linux
- [ ] test on RPi
- [ ] send related PR to oF main repo

 🎨 Update commandLinePG project files
Need to add vscodeProject.h/.cpp
- [x] Xcode
- [ ] visual studio
- [ ] Qt
- [ ] Code Blocks !?
 
⚠️ NOTICE
Project's template files must be located in [/openFrameworks/scripts/templates](https://github.com/openframeworks/openFrameworks/tree/master/scripts/templates)/vscode. I put these in this repo (/templates/vscode) in order to test quickly and deploy later. 

💡 vscode project configuration
I refer to following repos. Still not sure about the best settings.
https://github.com/npisanti/of_vscode
https://github.com/robertofazio/openFrameworks_VisualStudioCode_Example
https://github.com/hiroMTB/vscode_oF

 ✏️ Command line arguments
Please make sure to pass -p"vscode" argument when you debug command line executable.
In Xcode, we can set at Scheme.
-v
-o"/path/to/oF"
-a""
-p"vscode"
"/path/to/of/app/myApps/mySketch"

❗️Edit
- Please make a empty "vscode" folder under /openFrameworks/scritps/templates for frontend app test.

PR #203, opened 2019-05-21
strip ofTargetPlatform [experimental idea]

hiroMTB:
To be flexible to support another IDE, it might be better to remove ofTargetPlatform.
Because, for example, Qt, vscode, Atom, these IDE support multiple targets.

And for the name of the target like 'osx", ''vs', ''linuxarmv6l" etc, it would be nicer to use static const std::string, constexpr, or std::string_view etc instead of hardcoding with " ".

Related PR for vscode support -> https://github.com/openframeworks/projectGenerator/pull/202

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -56,7 +56,7 @@ std::string              directoryForRecursion;
 std::string              projectPath;
 std::string              ofPath;
 std::vector <std::string>     addons;
-std::vector <ofTargetPlatform>        targets;
+std::vector <std::string>     targets;
 std::string              ofPathEnv;
 std::string              currentWorkingDirectory;
 std::string              templateName;
@@ -87,7 +87,7 @@ bool printTemplates() {
     if(targets.size()>1){
 	std::vector<std::vector<baseProject::Template>> allPlatformsTemplates;
         for(auto & target: targets){
-            auto templates = getTargetProject(target)->listAvailableTemplates(getTargetString(target));
+            auto templates = getTargetProject(target)->listAvailableTemplates(target);
             allPlatformsTemplates.push_back(templates);
         }
 	std::set<baseProject::Template> commonTemplates;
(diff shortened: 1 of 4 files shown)


arturoc:
Yeah something alont this lines should work, the main problem i see is that it's really easy to mispell a platform name and you wouldn't get an error until run time if any. Perhaps something like this but using constants for the platforms would be enough.

We could even add a function to convert the ofTargetPlatform to string and then add new constants for other platforms inside the project generator addon?

Another possibility would be to leave the constants as they are but have strings to specify the IDE with defaults per platform, still in that case defining the strings as constants somewhere would be really useful to avoid errors from typos. If we had this there would be no immediate need to have to dropdowns yet if we can support vscode by now as an additional platform in the gui, but it would make it easier in the future to add support for  extra IDEs that are not supported on every platfom...

hiroMTB:
Thanks for your review arturo.
Yes, I will change raw string to constants in global scope.

Originally I was implementing to have another option for IDE, like you suggest, but somehow code got messy and change the plan. I will show the problem in another PR (or original vscode support PR).

I started think about that a fundamental problem is using ofTargetPlatform in PG.
For example, in current version, OF_TARGET_LINUX64 is associated to QtCreator.
But QtCreator project can be used in OSX and Windows too. (OF_TARGET_OSX, OF_TARGET_WINVS). So this PR is how it looks like if we remove ofTargetPlatform,

Also, I got new idea, which is similar with -t option. 
I found -t (--template) option in PG commandLine is quite useful and well designed.
Nice thing of -t option is we don't need to compile commandLine project. When we add a new folder in the /template directory, PG detects it. And we can use as an additional template to overwrite original template(like gl4.5, nofmod, etc). 

It might be nice to add this auto-detection feature to target template system as well. So when I add folder named /vscode,  /atom,  /emacs, /vim etc, let PG detect it automatically and create a dropdown list. 
Probably we need to have two separate directories such as /template/additionalTemplate and /template/targetTemplate in order to tell PG which is which.

arturoc:
Oh yeah you are right we can just use templates instead that should work without problem. Having a dropdown menu to select a template would be really useful.

arturoc:
This might happen from trying to render a really small rectangle in a slider which value is really small. If that's the case we could easily solve it by not drawing anything in sliders if the value is equivalent to less than one pixel which should be pretty cheap to calculate

hiroMTB:
Comment to another PR or forum post?

arturoc:
yeah thanks :)

hiroMTB:
Thanks, I will go for this route.
We don't need to do anything with PG for supporting vscode in the end.
Just need to put /template/vscode folder in oF main repo.
We can only generate vscode project via command line but I think it's a good starting point. 

I will check how to implement template dropdown later.

PR #205, opened 2019-05-24
Feature show hidden folder

hiroMTB:
Require to copy hidden .vscode folder


arturoc:
Can you change this to make files visible always instead of only for vscode? it's weird that there's any references from basproject to any specific project type but also it's pretty common to have dot files so better to always copy them

hiroMTB:
done :-)

arturoc:
thanks!

PR #206, opened 2019-05-25
Feature add template dropdown to frontend PG

hiroMTB:
# Add template dropdown list to Electron frontend PG
Screen Shot 2019-05-25 at 6 33 20 PM

## List of official template
```
var templates = {
    "emscripten": "Emscripten",
    "gitignore": "Git Ignore",
    "gles2": "Open GL ES 2",
    "gl3.1": "Open GL 3.1",
    "gl3.2": "Open GL 3.2",
    "gl3.3": "Open GL 3.3",
    "gl4.0": "Open GL 4.0",
    "gl4.1": "Open GL 4.1",
    "gl4.2": "Open GL 4.2",
    "gl4.3": "Open GL 4.3",
    "gl4.4": "Open GL 4.4",
    "gl4.5": "Open GL 4.5",
    "gles2": "Open GL ES 2",
    "linux": "Linux",  // !!??
    "msys2": "MSYS2/MinGW project template",
    "nofmod": "OSX application with no FMOD linking",
    "nowindow": "No window application",
    "tvOS": "Apple tvOS template",
    "unittest": "Unit test no window application",
    "vscode": "Visual Studio Code",
};
```

## Limitations and issues
- Only allow a user to select a single template.
- Platform and template combination can be false. For example, Linux64 + tvOS
- Some of the templates overwrite src files and main.cpp, some not. This can be problematic when a user just wants to update a project.
- We might not need gl3.1 ~ gl4.5 templates since it only has a small difference each other.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -202,7 +202,7 @@ position: fixed;
 						</div>
 						<div id="test"></div>
 					</div>
-					<div id="templatesSection" class="field">
+					<div id="templateSection" class="field">
 						<label>Template:</label>
 						<div class="ui multiple search selection dropdown" id="templatesDropdown">
 							<input name="templates" type="hidden">
@@ -258,7 +258,7 @@ position: fixed;
 							</div>
 						</div>
 					</div>
-					<div id="templatesSectionMulti" class="field">
+					<div id="templateSectionMulti" class="field">
 						<label>Template:</label>
 						<div class="ui multiple search selection dropdown" id="templatesDropdownMulti">
 							<input name="templates" type="hidden">
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -202,7 +202,7 @@ position: fixed;
 						</div>
 						<div id="test"></div>
 					</div>
-					<div class="field">
+					<div id="templatesSection" class="field">
 						<label>Template:</label>
 						<div class="ui multiple search selection dropdown" id="templatesDropdown">
 							<input name="templates" type="hidden">
@@ -258,7 +258,7 @@ position: fixed;
 							</div>
 						</div>
 					</div>
-					<div class="field">
+					<div id="templatesSectionMulti" class="field">
 						<label>Template:</label>
 						<div class="ui multiple search selection dropdown" id="templatesDropdownMulti">
 							<input name="templates" type="hidden">
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -204,7 +204,7 @@ position: fixed;
 					</div>
 					<div class="field">
 						<label>Template:</label>
-						<div class="ui search selection dropdown" id="templatesDropdown">
+						<div class="ui multiple search selection dropdown" id="templatesDropdown">
 							<input name="templates" type="hidden">
 							<i class="dropdown icon"></i>
 							<div class="default text">Template...</div>
@@ -260,7 +260,7 @@ position: fixed;
 					</div>
 					<div class="field">
 						<label>Template:</label>
-						<div class="ui search selection dropdown" id="templatesDropdownMulti">
+						<div class="ui multiple search selection dropdown" id="templatesDropdownMulti">
 							<input name="templates" type="hidden">
 							<i class="dropdown icon"></i>
 							<div class="default text">Template...</div>
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -202,6 +202,18 @@ position: fixed;
 						</div>
 						<div id="test"></div>
 					</div>
+					<div class="field">
+						<label>Template:</label>
+						<div class="ui search selection dropdown" id="templatesDropdown">
+							<input name="templates" type="hidden">
+							<i class="dropdown icon"></i>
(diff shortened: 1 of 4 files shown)


frontend/index.html
```diff
@@ -204,7 +204,7 @@
 					</div>
 					<div class="field">
 						<label>Template:</label>
-						<div class="ui search selection dropdown" id="templatesDropdown">
+						<div class="ui multiple search selection dropdown" id="templatesDropdown">
 							<input name="templates" type="hidden">
```
hiroMTB: notice : 
Use multiple dropdown UI here, but actually, it can only allow single template selection by using "maxSelections:1" option. See app.js, line 235, 257

arturoc:
This is really nice! I wonder if we could add some metadata to the templates to know for example which ones are compatible with the current platform or if they overwrite source files so they never appear when updating, only when creating a new project. I'll try to take a look later to see which flags would make sense

hiroMTB:
Thanks for review!
There is template.config file in each template folder. So we can check that file and only show valid template for selected platform. 

But if user select multiple platforms, I don't know what is the best behaviour of PG. 

For example, 
platform=osx, windows, 

Which template should be listed in the dropdown UI?

ofZach:
cool to see this!   do you think it would make sense to have the templates drop down visible only with "advanced" selected in the PG?  Just to keep things simpler for beginners....

hiroMTB:
True!

hiroMTB:
done!

arturoc:
About multiple platform selection, I think only templates that support all the selected platforms should be shown.

hiroMTB:
OK, I will work on it.

Current Commandline PG handles multiple platforms and wrong template correctly. It skips false combination and output warning.

```
$ projectGenerator -o"" -p"osx,vs" -t"nofmod" ./apps/test/testPlatformAndTemplate

[notice ] setting OF path to: /Users/hiroshimatoba/Desktop/openFrameworks/
[notice ] from PG_OF_PATH environment variable
[notice ] target platform is: osx
[notice ] using additional template nofmod
[notice ] updating project /Users/hiroshimatoba/Desktop/openFrameworks/apps/test/testPlatformAndTemplate
[notice ] parsing addons.make
[notice ] saving addons.make
[notice ] project updated! 

[notice ] setting OF path to: /Users/hiroshimatoba/Desktop/openFrameworks/
[notice ] from PG_OF_PATH environment variable
[notice ] using additional template nofmod
[notice ] updating project /Users/hiroshimatoba/Desktop/openFrameworks/apps/test/testPlatformAndTemplate
[warning] Cannot find nofmod using platform template only
[notice ] parsing addons.make
[notice ] saving addons.make
[notice ] project updated! 


1 project updated [notice ] in -6877.48 seconds
```

hiroMTB:
> About multiple platform selection, I think only templates that support all the selected platforms should be shown.

Updated!

Following image shows example behaviour when macOS and Windows are selected.
Template GL 4.2 is not supported by macOS so it is disabled (even Windows support it).

exsample-dropdown-with-multiple-platform

Please notice that tvOS folder does not contain a template.config file.
Also /templates/gitignore/template.config need to be modified to support ios and android I guess.

hiroMTB:
Ready to test and review.
So far I'm testing on macOS and works ok.

hiroMTB:
Rebased

arturoc:
Is it ready to merge then?

hiroMTB:
I am thinking to send another PR to add missing template.config to oF master.
But if you like Merge first and fix later, please merge :-)

hiroMTB:
Just tested on Windows too. 
I'm going to have a test on Linux but my Ubuntu does not boot  :-<
I will write again when it's ready.

hiroMTB:
I finished tset on Ubuntu 16.04. This is safe to merge I think.

Next step could be,
- restructure folder something like **/templates/platform** and  **/templates/project**
- synthesize template. e.g. **gitignore** + **gl4.1** + **vscode**
- show pop up alert when PG overwrite user file.
- cleaning up code
- update to ES6, 7, 8
- use Vue or React?
- integrate [ofPackageManager](https://github.com/thomasgeissl/ofPackageManager) by @thomasgeissl ?

p.s.
I had to do `apotheracy update poco` before execute compilePG.sh in both of latest master and nightly. Sometihng might be wrong with poco binary?

arturoc:
Thanks Hiroshi, this is a really nice addition!

About poco was that in linux?

NicTanghe:
Plz make the template options not hidden in advanced mode as it is impossible to know and add a setup guide for vscode (just for clarification that it is verry much supported)

hiroMTB:
I think vscode support is still experimental phase. And it might be difficult for beginners and students especially on Windows (since we need msys2). So better to hide as default. 

But it must be nice to have a setup guide. Do you think you can write a draft version while you are setting it up? Here is the .md files for another IDE as a reference. https://github.com/openframeworks/ofSite/tree/master/content/setup

PR #212, opened 2019-08-06
Implement ADDON_INCLUDES and ADDON_LIBS for linux64 / qtcreator

autr:
Parse headers & libs from ADDON_INCLUDES and ADDON_LIBS into generated .qbs project.

### DIFF
--- a/commandLine/commandLine.qbs
+++ b/commandLine/commandLine.qbs
@@ -6,10 +6,10 @@ import qbs.TextFile
 import "../../../libs/openFrameworksCompiled/project/qtcreator/ofApp.qbs" as ofApp
 
 Project{
-    property string of_root: '../../..'
+    property string of_root: "../../.."
 
     ofApp {
-        name: { return FileInfo.baseName(path) }
+        name: { return FileInfo.baseName(sourceDirectory) }
 
         files: [
             "src/main.cpp",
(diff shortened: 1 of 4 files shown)


arturoc:
I don't understand what this is doing, the addons in qtcreator are parsed by the qbs file, there's no need to parse them from the PG, just to include the name of the addon and the OF qbs build system  will do the same

autr:
On linux64 / Ubuntu 18.04 I found parameters set out inside each addon_config.mk are not being added to the .qbs file via the command line tool, so the project fails to link in cases where you are using dynamic or static libraries. From what I can see:

* [commandLine.qbs](https://github.com/openframeworks/projectGenerator/blob/master/commandLine/commandLine.qbs) is used as the template, but is missing `of.staticLibraries`, `of.frameworks`, `of.dynamicLibraries`. 
* [QtCreatorProject.h/cpp](https://github.com/openframeworks/projectGenerator/blob/master/ofxProjectGenerator/src/projects/qtcreatorproject.cpp) parses only the addon name (as opposed to VisualStudioProject, which processes all of the addon flags, libs, props, includes).
* [ofApp.qbs](https://github.com/openframeworks/openFrameworks/blob/master/libs/openFrameworksCompiled/project/qtcreator/ofApp.qbs) contains conditionals for `msys2` and `osx`, but no `linux64`.

autr:
Working example is these changes with my `addon_config.mk` / ofxZED addon:

https://github.com/autr/ofxZED/blob/master/addon_config.mk

Previously the command line PG would only successfully generate a Visual Studio project.

autr:
Working example is my `addon_config.mk` here:

https://github.com/autr/ofxZED/blob/master/addon_config.mk

I'm implementing CUDA 10.0 and ZED SDK libs, which would previously have to be manually added into the generated .qbs project.

arturoc:
This must be a problem in the qbs build files, as i explained above, with qbs we only need the name cause the qbs build system then parses the addon config itself and adds any necesary flags

commandLine.qbs is not a template but the project file used to build the project generator itself under qbs

I'm going to close this but feel free to keep posting if you have any doubts or opening a new issue explaining the original problem instead

PR #213, opened 2019-08-19
Bugfix adds linking for ofxPoco libs. Fixes nigthlies not building.

ofTheo:
This should fix the nightlies not building.  
The poco lib locations weren't being set in the makefile for the command line PG and so create_package.sh was failing when trying to compile the PG. 

NOTE: There might be another way to add this linking path.
This is what worked for me :)


arturoc:
this should be detected by the makefiles if the addon is there so it must be something else, 

i'll take a look when i have a moment but going to close this since hardcoding the library path like this is not correct.

ofTheo:
Ah yeah... totally makes sense.
Actually looking back at the commit history this was the commit where afterwards we stopped getting nightlies.  

https://github.com/openframeworks/openFrameworks/commit/e157173464d79cd604b0edb41764074461734190 

It might be as simple as just running install_dependencies.sh on the ci server? 

As poco got moved to being a system install for linux in that commit

PR #214, opened 2019-08-28
Multiple flags are correctly read into VS projects, added support for preprocessor definitions in addons

brinoausrino:
Before the project generator only saved the last flag of each type and overwrote the previous ones. This could be fixed by overwriting the first child of the note but not the not itself

from
`additionalOptions.first_child().set_value`
to
`additionalOptions.set_value((std::string(additionalOptions.value()) + " " + cflag).c_str());`

Addionally I added the support for preprocessor definitions in addons in VS projects.

### DIFF
--- a/ofxProjectGenerator/src/addons/ofAddon.cpp
+++ b/ofxProjectGenerator/src/addons/ofAddon.cpp
@@ -140,7 +140,7 @@ bool ofAddon::checkCorrectVariable(string variable, ConfigParseState state){
 				variable == "ADDON_DATA" ||
 				variable == "ADDON_LIBS_EXCLUDE" || variable == "ADDON_SOURCES_EXCLUDE" || variable == "ADDON_INCLUDES_EXCLUDE" ||
 				variable == "ADDON_DLLS_TO_COPY" ||
-				variable == "ADDON_DEFINES");
+				variable == "ADDON_DEFINES" || variable == "ADDON_PREPROCESSOR_DEFINITIONS");
 	case Unknown:
 	default:
 		return false;
@@ -261,8 +261,13 @@ void ofAddon::parseVariableValue(string variable, string value, bool addToValue,
 		addReplaceStringVector(includePaths,value,addonRelPath,addToValue);
 	}
 
+	
 	if(variable == "ADDON_CFLAGS"){
 		addReplaceStringVector(cflags,value,"",addToValue);
+	}
+
+	if (variable == "ADDON_PREPROCESSOR_DEFINITIONS") {
+		addReplaceStringVector(preprocessorDefinitions, value, "", addToValue);
 	}
 
 	if(variable == "ADDON_CPPFLAGS"){
@@ -310,7 +315,7 @@ void ofAddon::parseVariableValue(string variable, string value, bool addToValue,
 	}
 
 	if(variable == "ADDON_DATA"){
-		addReplaceStringVector(data,value,"",addToValue);
+		addReplaceStringVector(data,value,addonRelPath,addToValue);
 	}
 
 	if(variable == "ADDON_LIBS_EXCLUDE"){
--- a/ofxProjectGenerator/src/addons/ofAddon.h
+++ b/ofxProjectGenerator/src/addons/ofAddon.h
@@ -45,6 +45,7 @@ public:
     std::vector < std::string > frameworks;		// osx only
     std::vector < std::string > data;
 	std::vector < std::string > defines;
+	std::vector < std::string > preprocessorDefinitions; // vs only
 
     // metadata
     std::string name;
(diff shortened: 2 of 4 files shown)


arturoc:
the way overwriting works is by using += if you don't want to overwrite and just = if you want to overwrite, the syntax is based on makefiles syntax and allows to for example overwrite the defaults that are parsed from the file system which is quite useful for certain cases. or i'm understanding this wrong?

Also aren't preprocessir definitions the same as the already present ADDON_DEFINES?

brinoausrino:
Yes, this is how it should work. But the project generator for vs overwrites the setting even, when using +=. So I did the fix, that it works how it should.

Yes, you are right. I did not get it that the ADDON_DEFINES should be used for the preprocessor. The had a small bug as well. Fixed it and removed the separate preprocessor definitions

ofxProjectGenerator/src/addons/ofAddon.cpp
```diff
@@ -261,10 +261,15 @@ void ofAddon::parseVariableValue(string variable, string value, bool addToValue,
 		addReplaceStringVector(includePaths,value,addonRelPath,addToValue);
 	}
 
+	
 	if(variable == "ADDON_CFLAGS"){
 		addReplaceStringVector(cflags,value,"",addToValue);
 	}
 
+	if (variable == "ADDON_PREPROCESSOR_DEFINITIONS") {
```
arturoc: this reference to ADDON_PREPROCESSOR_DEFINITIONS should go as well

ofxProjectGenerator/src/addons/ofAddon.cpp
```diff
@@ -310,7 +315,7 @@ void ofAddon::parseVariableValue(string variable, string value, bool addToValue,
 	}
 
 	if(variable == "ADDON_DATA"){
-		addReplaceStringVector(data,value,"",addToValue);
+		addReplaceStringVector(data,value,addonRelPath,addToValue);
```
arturoc: what is this? doesn't seem related to any of the problems explained in the PR no?

ofxProjectGenerator/src/addons/ofAddon.h
```diff
@@ -45,6 +45,7 @@ class ofAddon {
     std::vector < std::string > frameworks;		// osx only
     std::vector < std::string > data;
 	std::vector < std::string > defines;
+	std::vector < std::string > preprocessorDefinitions; // vs only
```
arturoc: this should go as well

ofxProjectGenerator/src/addons/ofAddon.cpp
brinoausrino: No, sorry I don't really know why that happened.

ofxProjectGenerator/src/addons/ofAddon.cpp
brinoausrino: so, then adding ADDON_PREPROCESSOR_DEFINITIONS is useful?

Since I changed 
`additionalOptions = items[i].node().child("ClCompile").child("AdditionalOptions");`
for
`additionalOptions = items[i].node().child("ClCompile").child("PreprocessorDefinitions");`

Should we leave it like this or should we use the seperate preprocessor flag?

arturoc: no sorry, what i meant is it should go away

brinoausrino: already is in the latest commit

ofxProjectGenerator/src/addons/ofAddon.cpp
brinoausrino: ok, fixed it.

arturoc:
not sure what's going on but all the changes you've marked as resolved aren't really there. perhaps you forgot to push?

brinoausrino:
> 
> 
> not sure what's going on but all the changes you've marked as resolved aren't really there. perhaps you forgot to push?

Did I oversee something (do I need to make a new pull request, or does this work automatically)? The latest commit should resolve the addReplaceStringVector() thing:

https://github.com/openframeworks/projectGenerator/pull/214/commits/37b5cbfbe8342f4faca7ea5e5d9327c3c0544c49

The one before the preprocessor issue

https://github.com/openframeworks/projectGenerator/pull/214/commits/3ff92b5a1a22c3894f048ddcca7abdf0b747cbef

You have any idea?

arturoc:
those commits are already in, you don't need to create a new PR i think you are just missing some bits that are still referring to addon preprocessor define. look carefully at my annotations and the line numbers

brinoausrino:
sorry arturo, I was a little confused about the codeline. should be fixed now.

ofxProjectGenerator/src/addons/ofAddon.h
arturoc: this one is still missing, it should go away to. there should be no changes in ofAddon.h only on visualStudioProject

brinoausrino: excuse the inconvenience.

PR #221, opened 2019-10-16
adds _runMeFirst.command to macOS packages to escape translocation. closes #220

### LINKED ISSUE #220: Fix root path for macOS ( translocation / signing etc )
Moving this issue over here: https://github.com/openframeworks/openFrameworks/issues/6395 

@bakercp @ofZach is the signing approach something we can do right now? 

I tried quickly to code sign locally but it didn't solve the issue. 
Also took a look here which has some helpful info: https://github.com/electron/electron-osx-sign/wiki/3.-App-Sandbox-and-Entitlements

ofTheo:
Closes #220

Don't merge yet. 
Going to see if I can change the red dialog message to instruct users to run the command.

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -294,9 +294,9 @@ position: fixed;
 					</div>
 					<div class="ui negative message" id="ofPathSierraMessage" style="display: none">
 						<div class="header">
-							Running on OSX Sierra or later for the first time?
+							Running on macOS 10.12 or later for the first time?
 						</div>
-						<p>For security reasons, please set the OF path to root of openframeworks (ie, the folder which has "addons", "libs", in it)</p>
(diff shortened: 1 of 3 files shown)


ofTheo:
hmm the CI is failing due to glm errors on linux. 
https://travis-ci.org/openframeworks/projectGenerator/jobs/598897210 

might be it needs a day for the libs to be updated from the apothecary changes?

arturoc:
no, it seems the merge of the new glm has actually broken some things, master is broken at the moment https://github.com/openframeworks/openFrameworks

ofTheo:
@ofZach @arturoc - I just tried the PG with the xattr from within the PG fix ( #222 ) and it doesn't seem to solve the translocation.  So this might be the better approach for now.

ofTheo:
I think we should merge this in as a temporary fix until we figure out a better solution to the translocation issues. Going to restart the failed test and then merge.

PR #227, opened 2019-11-26
possible fix for linux pg errors. closes #209

### LINKED ISSUE #209: projectGenerator binary distributed with linux (and vs) is a MACH-O binary.
Thus, it won't run on Linux.

```
bakercp@fmachine:~/Programming/of/of_v20190324_linux64gcc6_release/projectGenerator-linux64/resources/app/app$ file projectGenerator 
projectGenerator: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|WEAK_DEFINES|BINDS_TO_WEAK|PIE|HAS_TLV_DESCRIPTORS>
```

The MACH-O binary is in these two at least  I haven't checked others.

https://openframeworks.cc/versions/v0.10.1/of_v0.10.1_linux64gcc6_release.tar.gz
https://openframeworks.cc/ci_server/versions/nightly/of_v20190324_linux64gcc6_nightly.tar.gz

They run on macOS if libmodex.dylib is present.

ofTheo:
Not able to test this, but this PR _should_ set the pgApp path to `apps/projectGenerator/commandline/bin/projectGenerator` for linux. 

Meaning that as long as the commandline PG is compiled it should run from the frontend. 

The other approach ( which it looks like is partly in the existing code ) is to call the projectGenerator from /usr/local/bin. But that requires the user installs it there first after compiling.  

This _should_ work in both cases as the PG has to be compiled at `apps/projectGenerator/commandline/bin/projectGenerator` first and is only copied when installed. So should exist at this location either way.

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -652,6 +652,7 @@ ipc.on('update', function(event, arg) {
     var templateString = "";
     var recursiveString = "";
     var verboseString = "";
+    var rootPath = defaultOfPath;
 
     if (update['updatePath'] !== null) {
         updatePath = update['updatePath'];
@@ -668,6 +669,7 @@ ipc.on('update', function(event, arg) {
 
     if (update['ofPath'] !== null) {
         pathString = "-o\"" + update['ofPath'] + "\"";
+        rootPath = update['ofPath'];
     }
 
     if (update['updateRecursive'] === true) {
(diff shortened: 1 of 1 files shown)


ofTheo:
Okay just tried this on a virtual Ubuntu install and it actually works. 
Sort of surprised, but before I apply this patch I get the projectGenerator missing error mentioned in #209 

Afterwards it generates and the project files fine. 
Going to push a fix for updating existing projects and then it should be good to go.

ofTheo:
okay this should be good now. 
👍

arturoc:
thanks!

PR #231, opened 2019-12-17
feature  ADDON_FRAMEWORKS_EXCLUDE to exclude frameworks from OSX projects

### LINKED ISSUE #230: add ADDON_FRAMEWORK_EXCLUDE
I'm currently running into a situation where I don't want to add all frameworks present in an addon by default.  

https://github.com/openframeworks/projectGenerator/blob/e11781ecd6af97f37187a39017587cb01c8280bd/ofxProjectGenerator/src/addons/ofAddon.cpp#L512

The library I'm wrapping ships several examples that contain 3rd party frameworks and I need to exclude them from the Xcode project with the PG.

danzeeeman:
This is a very simple update to fix #230. 
tested with [ofxVVISF](https://github.com/danzeeeman/ofxVVISF)


ofTheo:
Looks good to me. 
What is the use case for ofxVVISF just out of curiosity?

eduardfrigola:
Hi Theo, you can check here further discussion for ofxVVISF case: https://github.com/danzeeeman/ofxVVISF/issues/3

As there is an example folder inside the lib folder (which is a submodule of the library), that contains some example project with frameworks in it, there is no way to generate a project without those frameworks. Have I made myself clear?

ofTheo:
Thanks @eduardfrigola  :)

PR #242, opened 2020-04-19
Feature/version

thomasgeissl:
heyho,
i just added an option to get the current version.
this is quite helpful for me to integrate it with the new package manager frontend.

unfortunately v was already taken by the verbose option. it uses now (-w, --version). i am open for other suggestions.

mergin this pr, would also mean that we should start proper versioning for the projectGenerator. Not sure what the best option is.
* aligning with oF version, adding a counter for releases (e.g. 0.11.0-0)
* indepeneded versioning (e.g. 1.0.0)

let me know what you think.

unfortunately i had some code formatter running, so i guess this would break some PRs, if there are any related to the commandline.


arturoc:
the linux tests are failing with:

```
/home/travis/build/openframeworks/openFrameworks/apps/projectGenerator/commandLine/src/main.cpp: In function 'bool printVersion()':

/home/travis/build/openframeworks/openFrameworks/apps/projectGenerator/commandLine/src/main.cpp:97:1: error: no return statement in function returning non-void [-Werror=return-type]

 }
```

thomasgeissl:
i will fix that.
how about the versioning, any thoughts?

arturoc:
I would say independent versioning is better, perhaps we can track the major/minor for OF but have separate patch 

so for OF 0.11.0 we would release PG 0.11.0 but if there's a patch release for OF it would be 0.11.1 and if the PG doesn't have any changes it would still be '0.11.0

thomasgeissl:
i think the hybrid versioning would be confusing. pg 0.11.1 would make me asume that there is a corresponding of 0.11.1.
i am fine starting versioning with 1.0.0.
i guess #243 would be then version 2 already.

arturoc:
i think it's easier to track if we use the same major/minor for both PG and OF. if someone thinks that 0.11.1 has to be used with 0.11.1 that's not so problematic i think and can be documented but having a way to easily identify which version of the PG can be used with each version of OF without having to check out some document everytime could be useful

Not sure it's super important since most versions just work with older versions of OF but still

thomasgeissl:
i changed the version to 0.11.0.

if this gets merged, would it possible to tag it and upload the build the github releases? that would things easier for me to make it installable via brew and hopefully chocolatey on windows.

thomasgeissl:
hey @arturoc, any chance to get this merged?

ofTheo:
@thomasgeissl the functional changes look good to me but the PR also includes a whole bunch of formatting changes. Could you resubmit with just the changes relating to versioning?

thomasgeissl:
i submitted another PR.

Issue #243, opened 2020-04-19
cross platform cli api


thomasgeissl:
hello,
it would be very nice to have a cross platform cli api. (e.g. -o instead of \o on windows)
i could add cxxopts or any other option parser.
but i am not sure where it would break things. i guess the pg is currently used to create the platform packages, is not it?
i am happy to make this change if you are fine with it.
thomas

arturoc:
yeah the options parser is not very good and it has other problems so having a better one would be useful i think

thomasgeissl:
i had good experiences with cxxopts. i try to get it done this week.

thomasgeissl:
https://github.com/thomasgeissl/ofProjectGenerator/tree/refactor/cliarguments

i still need to test and update the readme. but in case someone wantes to give feedback already, here it is.

PR #245, opened 2020-04-27
Refactor/cliarguments

thomasgeissl:
i replaced the option parser with cxxopts, it should have the interface now on all platforms.

* the option flags changed a bit. 
  * the project path needs now the p flag instead of appending it at the end (i havent found a way to do it without a flag). 
  * platforms is now -s or --platforms
* options containing more than one word need to be wrapped in double quotes

```
#!/bin/sh
cd "$(dirname "$0")"
PG=./bin/projectGenerator.app/Contents/MacOS/projectGenerator 
OF=/Users/thomas.geissl/libs/of_v0.11.0_osx_release/

$PG -h
$PG --help

$PG -l -o $OF


mkdir test

$PG -o $OF -p test/first
$PG -o ../../.. -p test/second -a ofxOsc,ofOpenCv
$PG -o ../../.. -p test/second -a "ofxOsc, ofxOpenCv, ofxSvg"
$PG -o ../../.. -p test/second -a "ofxOsc, ofxOpenCv, ofxSvg" -s osx -v

$PG -o ../../.. -p test -r -d -v
PG_OF_PATH=../../.. $PG -p test/third -a "ofxOsc, ofxOpenCv, ofxSvg" -t nofmod

```

the version flag is still missing, but in case you wanna merge this, i think it is important to first merge my other pr #242 and tag it.
then this would be already a new major release.

i only tested on osx.

let me know what you think, anything i should improve?


thomasgeissl:
i totally forget the frontend, will do the changes this week.

thomasgeissl:
turned out that the cli readme is outdated and the option parser works already the same on all platforms.
https://github.com/openframeworks/projectGenerator/tree/master/commandLine

i still think it makes sense to replace it with another more intiutive option parser.
https://github.com/openframeworks/projectGenerator/issues/243

dimitre:
I think this is a good change. But we have to change the way it handle parameters right? like -o"../../.." needs a space between -o and value ?

PR #248, opened 2020-06-29
feature -  ability to specify additional source/include folders to add to project

ofTheo:
This is just the command line PG changes necessary to add other source / includes folders to a project outside of the project directory.  

PR adds  a `--source` or `-s` argument to the command line PG and lets you specify both absolute or relative source or include paths to add to the project.  eg: 

```
/projectGenerator -o"../../../../" -s"../../../../SharedCode" -p"osx"  ../../../../apps/myApps/testPG1
```

This is super helpful for Visual Studio users, as each time you regenerate a project ( if you need to include another addon ) you lose any other files and folders added and unlike other IDEs it is very tedious to add back in any shared code folders each time. 

Before I start working on the frontend changes I just wanted to get some thoughts on the best way to cache/store the added source paths so they can be loaded back in by the frontend. 

Right now we use addons.make to keep track of the addons added to a project. 
I was hoping to avoid using another file to keep track of the file paths. 

config.make has `PROJECT_EXTERNAL_SOURCE_PATHS` which has the same idea / functionality.  
Would that make sense or be weird?

With any PG changes we might be thinking about is there a need for a file to store additional info for the frontend at a per project level? Or is config.make the best solutions for this?  

Would love any thoughts @arturoc @thomasgeissl 

ps: I realize @thomasgeissl #245 uses the -s argument in a different way. Happy to coordinate so we don't end up using the same args. :)


ofTheo:
Tested now on Visual Studio and Xcode. 
Seems to work well for both absolute file paths and relative file paths.

ofTheo:
going to merge this. we've been using it now for a while on mac and pc and its been a huge help ( especially for PC ).

PR #249, opened 2020-08-05
added version option

thomasgeissl:
hey @ofTheo,
here without the formatting.


thomasgeissl:
here for reference the old PR, with formatting accidentally turned on.
https://github.com/openframeworks/projectGenerator/pull/242

ofTheo:
thanks @thomasgeissl !

thomasgeissl:
probably it makes sense to display the version on the frontend.

@ofTheo would be possible to tag the release then and upload a signed version to the github releases or the oF ci servers. That would make it much easier to automate things for me with the package manager, e.g. make it installable via brew.

ofTheo:
@thomasgeissl I think the versions used for releases are automatically archived here: 
http://ci.openframeworks.cc/projectGenerator/ 

It might be possible to automatically add them to the GitHub releases via their api but I haven't tried to do that yet. 
Would the above link work for your needs? I believe the osx PG app is signed via the CI.

thomasgeissl:
that should work. 
but now the versions do not necessarily match with the oF versions as we discussed here.
https://github.com/openframeworks/projectGenerator/pull/242#issuecomment-618935275

i guess compiling and signing on travis and automatically deploy if a commit is tagged is too much work for now. but i think we should at least tag the releases on github, then we can use features like the rss feed.
https://github.com/openframeworks/projectGenerator/releases.atom

PR #251, opened 2020-09-30
Fix multiple level dependencies, cache disk access to speed up generation.

armadillu:
This PR fixes an issue where the order in which addons are supplied can lead to different results, with addons sometimes not being included in the project. This only affects addons that are not stored in the `OpenFrameworks/addons` directory.

On projects where local addons are included, if those local addons have other local addons as a dependency, it could lead to a situation where some of those local addons end up not being added to the project. 

This big has been biting me for a very long time; I finally took the time to setup the CommandLine Xcode Project to be able to debug and find the issue. It turns out that PG keeps a list of addons that have been added to the project; if an addon has already been added, it will skip any other requests to add it. The problem comes when this first addon is added but it's empty. It shows up on the list as added, but nothing was really added, as the folder itself wasn't found. This can happen when PG parses dependcies of a local addon; I guess it expects all the dependencies to be "native" addons, but that's not always the case. Lots of local addons depend on other local addons, but PG currently just looks for depedency addons in the `OpenFrameworks/addons` folder. 

With the current implementation, if a folder does not exist with the addon name in `OpenFrameworks/addons`, PG still assumes its a valid addon, adds it to the project (which leads to nothing, as the addon has 0 files, 0 libs, etc) and puts it in the list of "added addons". This means if that later on it gets a request to add this same local addon but with the correct path (ie ../../externalAddons/ofxAnimatable), it would add the addon correctly, but bc ofxAnimatable is already on the list of added addons, its rejected and ignored... And the addon ends up not being part of the final created project.

The core of the fix is to not add addons to the projects if the folder is just not there. 

By making `fromFS()` return a bool we can indicate if it succeeded or failed to load an addon from disk. If it failed, then it was never an addon to begin with, so it should not take a spot on the "added addons" list.

Another feature of this addon is caching the result of `fromFS()` to RAM. This method is quite slow; for example ofxOpenCV takes several seconds. Currently, if multiple addons have it as a dependency, its scanned from disk every time its listed as a dependency, leading to 90+ seconds to create a project in some cases. With this PR, if an addon is loaded/indexed from disk once and its valid, it's cached to RAM so the next time it's requested it can skip the io access.

This leads to 3x faster generation on my test setup.

I tested this on windows and osx.

PS.

This PR doesn't fix the issue where local addons that have other local addons as a dependency will not get those dependencies added. Currently, "addon_config.mk" is parsed for dependencies, and for each addon listed there, it only looks inside the `OpenFrameworks/addons` folder for it. As local addons are often not there but in some other folder, they are not found this not added to the project. 

I usually keep addons that are not part of the OF release besides the OpenFrameworks folder, in a folder named "ExternalAddons". To create a project using those addons, I supply a relative path and it works great.

```
 ./projectGenerator -o"/Users/oriol/Desktop/OF_uricat/OpenFrameworks" -p"osx" -a"ofxOsc, ofxXmlSettings, ofxPoco,../../ExternalAddons/ofxAnimatable" ...
```

This is kind of hard to fix; as if ofxAnimatable has a dependency that's in "../../ExternalAddons", with the current API, PG has no way of knowing that... Unless it does some guesswork. We could look at all the paths that have been supplied for addons so far and dynamocally learn where to look for them; but I want to see what you think first. Another option is adding more CLI options to supply alternative addon locations (paths) in which to look for dependencies.

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -1,7 +1,8 @@
 #include "ofMain.h"
 #include "optionparser.h"
 #include "defines.h"
-enum  optionIndex { UNKNOWN, HELP, PLUS, RECURSIVE, LISTTEMPLATES, PLATFORMS, ADDONS, OFPATH, VERBOSE, TEMPLATE, DRYRUN, VERSION };
+enum  optionIndex { UNKNOWN, HELP, PLUS, RECURSIVE, LISTTEMPLATES, PLATFORMS, ADDONS, OFPATH, VERBOSE, TEMPLATE, DRYRUN, SRCEXTERNAL, VERSION};
+
 constexpr option::Descriptor usage[] =
 {
     {UNKNOWN, 0, "", "",option::Arg::None, "Options:\n" },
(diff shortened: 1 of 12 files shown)


arturoc:
seems like the merge of the last commit has broken this, can you rebase it?

armadillu:
ok! Having a look now...

armadillu:
nope! I botched that, I will send another PR.

armadillu:
I botched this, sending another PR.

armadillu:
ok look here instead https://github.com/openframeworks/projectGenerator/pull/252

PR #286, opened 2022-03-09
Fix ADDON_C_SOURCES is ignored.

2bbb:
This PR fixed bug ADDON_C_SOURCES is ignored.
with replacing too long connection likes `variable == "VARIABLE" || variable == "VARIABLE"  || variable == "VARIABLE" || ...` to declare vector of variable and using `std::find`.

This update makes be clear declarations by `std::vector<std::string>` at the top of the header file. 
I think it is good as a kind of self-documentation.

And replace magic number like "VARIABLE_NAME" with std::string VARIABLE_NAME = "VARIABLE_NAME".
Those strings are used only twice but I believe to making constant variables guarantee the same value.


ofTheo:
@2bbb - sorry I missed this. 
Any chance you could update and we can merge it?

ofTheo:
Thank you @2bbb !

PR #290, opened 2022-04-13
change templatesFolder hardcoded to a variable

dimitre:
if we decide to change templates folder to root as suggested


ofZach:
can you link back to the discussion of moving templates to root?   I was just going to chime in with a suggestion that I think we don't users to interact with templates accidently so it makes sense to have them "removed" from the root in some way.  It's weird to me that they are in a folder called scripts (this is more historical than anything else) but maybe there's an organization that makes more sense but also shield the user from accidentally opening up a template, altering it, and hosing their whole OF folder.

dimitre:
Sure, it was discussed here
https://github.com/openframeworks/openFrameworks/issues/6866#issuecomment-1057054083

PR #294, opened 2022-05-29
Github Actions - update linux

dimitre:
service not available anymore on 16.04

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -10,7 +10,7 @@ on:
     
 jobs:
   build-linux:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     strategy:
       matrix:
         cfg:


dimitre:
@ofTheo @danoli3 can you please merge this one?
16.04 was discontinued recently
https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/

ofTheo:
Hmm, looking into this more I think the issue is that the gcc that is default in Ubuntu 18.04 still has the experimental filesystem ( https://askubuntu.com/questions/1365157/upgrade-gcc-version-in-ubuntu-18-04 ). So we might need to change the way we are passing flags to the compiler. 

I'll try and fix this on the main OF repo, as OF should compile out of the box on 18-04. Once I think its working we can rerun the check here and then merge.

dimitre:
Great! what about 20.04?

ofTheo:
This works now with https://github.com/openframeworks/openFrameworks/pull/7013 merged

Issue #297, opened 2022-07-21
projectGenerator CLI creates an empty addons.make where it is not needed


dimitre:
I'm opening this issue which is a sister of this other one https://github.com/openframeworks/projectGenerator/issues/258
I'll probably be trying to fix it. 
Reason: you have changes in git if you run scripts to build all examples. new empty files created where not needed.

PR #298, opened 2022-07-24
Avoid creating an empty addons.make

dimitre:
Avoid creating an empty addons.make
if there are no used addons.

closes https://github.com/openframeworks/projectGenerator/issues/297

### DIFF
--- a/ofxProjectGenerator/src/projects/baseProject.cpp
+++ b/ofxProjectGenerator/src/projects/baseProject.cpp
@@ -224,15 +224,16 @@ bool baseProject::create(string path, std::string templateName){
 
 bool baseProject::save(){
     ofLog(OF_LOG_NOTICE) << "saving addons.make";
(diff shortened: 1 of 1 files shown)


ofZach:
is there a practical reason for this change?  The reason I ask is that I find it helpful to look at the addons.make almost a kind of list of addons, and having an empty one gives me some insight into a project.   I realize that not having an addons.make would also give some info, but it could also be a missing file so it feels a little more ambiguous.   I saw this note 

> you have changes in git if you run scripts to build all examples. new empty files created where not needed.

but not sure I follow...  I'm definitely not opposed to the change but was trying to understand it a little better.

dimitre:
Hello @ofZach thanks for the opinion. 
No, there is no special practical reason
just sounded as "more correct" to me, but I suppose it is personal opinion. I'm ok with it as it is too.
for me it is like an empty .gitignore file. You can have it and it can declare an intention but I would rather not having any if not needed.

PR #301, opened 2022-07-27
Project rename fix

dimitre:
I missed something in the big Json PR, and just noticed all apps were emptyExample
so I've added the findandreplaceInTexfile again for the project.pbxproj


ofTheo:
cool - just running the checks and will merge if all good! 👍

PR #305, opened 2022-09-09
fix recursive addons issue

dimitre:
closes https://github.com/openframeworks/openFrameworks/issues/7067


ofTheo:
Awesome - thanks @dimitre 

Such a small fix - but makes sense why it would break the recursive process. 
Static vars in functions can be super weird :)

PR #307, opened 2022-09-21
Json automatic conversion from string to array when needed

dimitre:
This PR addresses issues with new versions of XCode which converts one array with one item to a simple string.
Here we save this value (if it is not null) create an array in place, push the old value back, and the new value in the array.
Tested with complex addons like ofxSyphon.

Good to merge together with this one https://github.com/openframeworks/openFrameworks/pull/7084


ofTheo:
Thanks @dimitre !

PR #310, opened 2022-10-01
Fix Pugi errors

dimitre:
updating pugi functions to non-deprecated ones.


dimitre:
@ofTheo this is good to merge and replaces deprecated functions of pugixml

PR #315, opened 2022-12-15
ios template fix

dimitre:
to fix the issue mentioned here
https://forum.openframeworks.cc/t/ofxpoco-not-working-with-nightly-builds/40916

to be merged together with this
https://github.com/openframeworks/openFrameworks/pull/7252


ofTheo:
Thanks so much  @dimitre! 

Going to give it a few mins so the PG can get uploaded and then will merge https://github.com/openframeworks/openFrameworks/pull/7252

PR #318, opened 2023-03-07
of::filesystem update

dimitre:


dimitre:
@ofTheo this is ready to go.
Other than update PG to the almost minimum changes to run with of::filesystem I've updated github actions, macOS to latest, actions now will cancel under new commits, and I had to update buildPG.sh to update npm and fix permissions to work OK on runners.

ofTheo:
For future stuff we should prob move more of the std::string to of::filesystem::path and avoid converting to string as much as possible so that we could better support non US/EN filesystems. Imagine that would be more of an undertaking though. :)

dimitre:
Yeah indeed! lots of strings and vectors, we can get rid of hardcoded path separators too

PR #322, opened 2023-03-10
PG of::filesystem fix

dimitre:
Still not a complete PR, but my intention is to finish it today


dimitre:
Yeah. This was PG related issues, and my latest PR, it seems I didn't test enough.
Now I've ported some other parts to rely on filesystem to split paths and do path operations. 
It should already work OK now.

PR #332, opened 2023-04-04
Fix paths for projects when not relative to OF folder

dimitre:
I think it closes https://github.com/openframeworks/openFrameworks/issues/7464

### DIFF
--- a/ofxProjectGenerator/src/projects/xcodeProject.cpp
+++ b/ofxProjectGenerator/src/projects/xcodeProject.cpp
@@ -129,8 +129,11 @@ bool xcodeProject::createProjectFile(){
 	// make everything relative the right way.
 	string relRoot = getOFRelPathFS(projectDir).string();
 	projectDir = projectDir.lexically_normal();
-	if (projectDir.is_absolute()) {
-		relRoot = getOFRoot();
+        
+        //projectDir is always absolute at the moment
+        //so lets check if the projectDir is inside the OF folder - if it is not make the OF path absolute
+	if( projectDir.string().rfind(getOFRoot(),0) != 0 ){
+            relRoot = getOFRoot();
 	}
 	if (relRoot != "../../.."){
 		findandreplaceInTexfile(projectDir / (projectName + ".xcodeproj/project.pbxproj"), "../../..", relRoot);
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -14,8 +14,7 @@ concurrency:
     
 jobs:
   build-linux:
-    runs-on: ubuntu-18.04
-    # runs-on: ubuntu-latest
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         cfg:
@@ -23,6 +22,8 @@ jobs:
     env:
      TARGET: ${{matrix.cfg.target}}
     steps:
+    - name: Install libunwind
+      run:  sudo apt-get install libunwind-dev
     - uses: actions/checkout@v3
     - name: Build
       run:
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -20,7 +20,7 @@
     "build:osx": "electron-packager . $npm_package_productName --icon=static/icon/of.icns --out=dist --ignore='dist' --ignore='readme.md' --prune --overwrite --electron-version=0.30.0 --platform=darwin --arch=x64",
     "build:linux32": "electron-packager . $npm_package_productName --icon=static/icon/of.icns --out=dist --ignore='dist' --ignore='readme.md' --prune --overwrite --electron-version=0.30.0 --platform=linux --arch=ia32",
     "build:linux64": "electron-packager . $npm_package_productName --icon=static/icon/of.icns --out=dist --ignore='dist' --ignore='readme.md' --prune --overwrite --electron-version=0.30.0 --platform=linux --arch=x64",
-    "build:vs": "electron-packager . --icon=static/icon/of.ico --out=dist --ignore='dist' --ignore='readme.md' --prune --overwrite --electron-version=0.30.0 --platform=win32 --arch=ia32"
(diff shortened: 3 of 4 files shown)


ofTheo:
Ahh - thanks @dimitre! 
One thing I noticed when testing #332 is projectDir was always absolute for me ( at least at this point https://github.com/openframeworks/projectGenerator/blob/master/ofxProjectGenerator/src/projects/baseProject.cpp#L22 )

So if I passed in a relative path for the project or an absolute path or an absolute path outside of the OF path `projectDir` was always absolute at that point. Things get made relative later but just good to know especially if you are relying on `projectDir` being both relative and absolute ( which it doesn't seem to be ). 

Not sure if that affects this PR. But just wanted to mention it. 

Also wondering how many filesystem bugs we might fix ( for spaces and non English characters ) if we handled paths relatively from an early stage. In the PG app.  ( ie: if OF is a relative path and project is relative, we don't even need an abs path ). But that might be another large PR / project 🙂

ofxProjectGenerator/src/projects/xcodeProject.cpp
```diff
@@ -128,10 +128,12 @@ bool xcodeProject::createProjectFile(){
 
 	// make everything relative the right way.
 	string relRoot = getOFRelPathFS(projectDir).string();
-
+	projectDir = projectDir.lexically_normal();
+	if (projectDir.is_absolute()) {
+		relRoot = getOFRoot();
```
ofTheo, line 133: my experience is getOFRoot() is always absolute - but might be worth checking.

dimitre: I think it is, yes. in this case I'm setting relRoot to absolute root path, so we don't have some absurd relative paths going all the way down to root, like
../../../../../../../Users/z/Desktop/project

ofTheo: I think you might need a bit more handling here.  First you would want to make `relRoot` relative to projectDir but  only make relative if the projectDir exists inside the OF folder. 

I think very similar to what I did here: 
https://github.com/openframeworks/projectGenerator/blob/master/ofxProjectGenerator/src/projects/baseProject.cpp#L221-L224

dimitre: obviously in this case relRoot is misnamed. in fact updating PG always have the feeling it has to be completely rewritten :)

ofTheo: Otherwise I think it will end up replacing ../../../ with absolute OF paths :)

ofTheo: oooops - nevermind - I misread the changes. 
relRoot starts relative and is made absolute only if projectDir is absolute. 

I think in this case you would actually want to just check if the projectDir is not inside the OF folder then do: 
`relRoot = getOFRoot();` 

( so the opposite of this - https://github.com/openframeworks/projectGenerator/blob/master/ofxProjectGenerator/src/projects/baseProject.cpp#L222 ) 

Sorry for the noise!! 🤦‍♂️

dimitre: yeah this is the best option.

ofTheo: cool - will do! 
I just checked and it is currently making projectDir absolute - so then then relRoot gets made abs. 

```
 relRoot is ../..
 projectDir is "/Users/theo/Documents/CODE/__OPENFRAMEWORKS/openFrameworks/apps/test123"
 relRoot is /Users/theo/Documents/CODE/__OPENFRAMEWORKS/openFrameworks
```

dimitre: yeah with your path inside path detection it will do it right
yesterday before your PR I was searching for the same thing. I was wondering if std::filesystem had a function to check if a path contains another, so nothing in this case. 
I was going to test this snipped but I've seen you already solved in your PR
Maybe we could have a function in Utils.cpp (or the core) so your solution can be reused if needed in other places.

```c++
bool is_subpath(const std::filesystem::path &path,
                const std::filesystem::path &base)
{
    auto rel = std::filesystem::relative(path, base);
    return !rel.empty() && rel.native()[0] != '.';
}
```

ofTheo:
Okay this should be good now. 

a few tests: 

```
[notice ] setting up new project /Users/theo/Documents/CODE/__OPENFRAMEWORKS/openFrameworks/apps/projectGenerator/commandLine/bin/../../../../apps/test345
 relRoot is ../..
 projectDir is "/Users/theo/Documents/CODE/__OPENFRAMEWORKS/openFrameworks/apps/test345"
 relRoot is ../..
```

```
[notice ] setting up new project /Users/theo/Documents/CODE/__OPENFRAMEWORKS/openFrameworks/apps/projectGenerator/commandLine/bin/../../../../apps/myApps/123App
 relRoot is ../../..
 projectDir is "/Users/theo/Documents/CODE/__OPENFRAMEWORKS/openFrameworks/apps/myApps/123App"
 relRoot is ../../..
```

```
[notice ] setting up new project /Users/theo/Desktop/_Test345
 relRoot is ../../Documents/CODE/__OPENFRAMEWORKS/openFrameworks
 projectDir is "/Users/theo/Desktop/_Test345"
 relRoot is /Users/theo/Documents/CODE/__OPENFRAMEWORKS/openFrameworks
```

And all the OF files show up in the side bar as expected.

ofxProjectGenerator/src/projects/xcodeProject.cpp
ofTheo: @dimitre yes! great idea. having a `is_subpath` or even more specific `isPathInOFRoot` or something would be helpful - the rfind approach is kind of hard to follow ( though it seems to work well )

PR #333, opened 2023-04-06
fix config.make to use relative OF_ROOT if project is inside OF folder

ofTheo:
This addresses the issue here:
https://forum.openframeworks.cc/t/using-of-with-vscode-on-windows/41676/2 

And probably will address some other issues where examples work but projects don't. 
It makes OF_ROOT relative if the project exists inside the OF folder. 

ie for most projects in `apps/myapps/` OF_ROOT will now be `../../../` - currently it was hard coding the full absolute path which means if you moved your OF folder the project would break. 

Wondering if there might be other instances ( in other project file generation ) where we should be using relative paths and we are using absolute.

### DIFF
--- a/ofxProjectGenerator/src/projects/baseProject.cpp
+++ b/ofxProjectGenerator/src/projects/baseProject.cpp
@@ -217,7 +217,13 @@ bool baseProject::save(){
 
 			//add the of root path
 			if( str.rfind("# OF_ROOT =", 0) == 0 ){
-				saveConfig << "OF_ROOT = " + getOFRoot() << endl;
+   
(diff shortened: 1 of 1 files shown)


dimitre:
Great one Theo, this was a much needed fix.

dimitre:
I don't remember other cases of this need. but this one is a little bit similar
https://github.com/openframeworks/projectGenerator/pull/332

I decided to check if projectDir is absolute and use absolute paths or not based on this. this can fix the issue with generating projects in other directories outside of openFrameworks.
But it can be updated to use the check if the OF path is contained in project path, use relative. if not use absolute instead

PR #335, opened 2023-04-07
XCode - addons with correct paths now

dimitre:
In XCode if you click any addon folder, now it has the correct location.
if you right click -> show in finder it will open the correct folder.

And it won't create a path if it is a framework, so we don't have a broken link there.

EDIT: this also fix one issue about PG not generating other templates like vscode.


ofTheo:
awesome! I was wondering about the template part - then saw the note 🙂

PR #339, opened 2023-04-12
projectGenerator local_addon fixes

dimitre:
This is a fix of local_addons functionality
translates some ofDirectory operations and listings to fs::path, so it removes the complexity of using ofToDataPath
because projectGenerator/bin/data is not used.

with this PR we can remove "local_addons" folder from the project itself, it will be created from PG


ofTheo:
one thing to look at is the non Xcode projects/files that reference `ofxProjectGenerator`: 
https://github.com/openframeworks/projectGenerator/search?q=ofxProjectGenerator 

Ideally this PR would also update those files so they continue to be able to compile the PG.

ofTheo:
ahh thanks - already did it via:
https://github.com/openframeworks/openFrameworks/commit/a76ef83f0a6eaf1079fbe35a444ae721225e5fee  

slightly different approach - but should work

dimitre:
I'm also removing Poco manually from everything. :) 🧹

dimitre:
Finally, cleanup done and passing tests.
I just don't have how to test if the project files edits are ok

PR #343, opened 2023-04-13
Frameworks path in local_addons fix

dimitre:
Fixed a hard coded local_addons when adding Frameworks from addons.

Fix a issue where addons were added multiple times when added by dependencies of other addons. The result is the same but it is more performant.

Update some functions to fs::path

dylibs files now goes to its own folder in local_addons too.

And some minor changes.


dimitre:
I'm about to finish this one, with my tests here it is about 100 times faster to build complex projects.
from 
1 project updated [notice ] in 4.9661 seconds
to 
1 project updated [notice ] in 0.0535292 seconds

dimitre:
Hello @ofTheo this one is good to go. 
There are plenty of much needed updates. 
I'm now making lots of tests with different addons and platforms. 
BTW, dylibs are working great !

ofTheo:
great! feel free to merge when ready!! 🎉🙂

Issue #342, opened 2023-04-13
version of electron environment is too old.


2bbb:
I tried to build frontend, but `npm i` failed with `electron-prebuilt` that search too old electron binary zip and not found for darwin-arm64.
and I removed it from devDependencies and build again, then it got success to build.

electron-prebuilt was marked deprecate long long year ago.
can we remove this (or replace similar one if needed)?

2bbb:
Error log:

```
2bit@2bM1Max frontend % npm i
npm WARN old lockfile
npm WARN old lockfile The npm-shrinkwrap.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
npm WARN deprecated xmldom@0.1.31: Deprecated due to CVE-2021-21366 resolved in 0.5.0
npm WARN deprecated uuid@3.3.3: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
npm WARN deprecated request@2.88.0: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated ini@1.3.5: Please update to ini >=1.3.6 to avoid a prototype pollution issue
npm WARN deprecated har-validator@5.1.3: this library is no longer supported
npm WARN deprecated electron-osx-sign@0.4.17: Please use @electron/osx-sign moving forward. Be aware the API is slightly different
npm WARN deprecated asar@0.14.6: Please use @electron/asar moving forward.  There is no API change, just a package name change
npm WARN deprecated electron-download@4.1.1: Please use @electron/get moving forward.
npm WARN deprecated mkdirp@0.5.1: Legacy versions of mkdirp are no longer supported. Please update to mkdirp 1.x. (Note that the API surface has changed to use Promises in 1.x.)
npm WARN deprecated electron-download@1.4.1: Please use @electron/get moving forward.
npm notice
npm notice New major version of npm available! 8.15.1 -> 9.6.4
npm notice Changelog: https://github.com/npm/cli/releases/tag/v9.6.4
npm notice Run npm install -g npm@9.6.4 to update!
npm notice
npm ERR! code 1
npm ERR! path /Users/2bit/prog/of/openFrameworks/apps/projectGenerator/frontend/node_modules/electron-prebuilt
[...]

PR #345, opened 2023-04-15
switch electron to 0.37.8 to temp fix broken 0.30.0

ofTheo:
0.30.0 is erring when fetching as mentioned in #342 
This doesn't solve the arm64 issue or the fact that electron-prebuilt is super out of date but it does fix immediate issues. 

We still need to bring electron up to date, but that means a rewrite of index.js and app.js 

@2bbb curious if these changes work for you?

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -157,11 +157,8 @@ var mainWindow = null;
 
 // Quit when all windows are closed.
 app.on('window-all-closed', function() {
-    // On OS X it is common for applications and their menu bar
-    // to stay active until the user quits explicitly with Cmd + Q
-    if (process.platform != 'darwin') {
-        app.quit();
-    }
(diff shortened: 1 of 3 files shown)


2bbb:
@ofTheo 
sorry, i don't check.
but maybe 0.xx is too old...

i run with newest electron, and i don't know why some part is worked, some features is not worked...
and i test with npm run (i.e. electron run) then it failed to run...

main problems of about updating to newer electron is around ipc.
and i will try to replacing ipc system to newer one.

ofTheo:
> main problems of about updating to newer electron is around ipc.
> and i will try to replacing ipc system to newer one.
> 

@2bbb that would be amazing. 🙏
Would be great to get the code / electron version up to date.

Issue #348, opened 2023-04-18
Actions, cache Linux too


dimitre:
in the likes of macOS

PR #354, opened 2023-04-19
Linux Caching 

dimitre:
caching OF repo cloned inside the scripts here.
same as macOS action.

closes https://github.com/openframeworks/projectGenerator/issues/348


dimitre:
No it doesn't. I think it can be useful if we update buildPG.sh to pull if directory already exists.

PR #355, opened 2023-04-19
Create / Update

dimitre:
This PR makes recursive file update outside the create / update block,
fix issue with projectName being treated differently for paths with trailing slashes or without
Adds new archs for RPI 4

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -148,6 +148,9 @@ void addPlatforms(const std::string & value) {
 		else if (p == "linuxarmv7l") {
 			targets.emplace_back(OF_TARGET_LINUXARMV7L);
 		}
+		else if (p == "linuxaarch64") {
+			targets.emplace_back(OF_TARGET_LINUXAARCH64);
+		}
 		else if (p == "msys2") {
 			targets.emplace_back(OF_TARGET_MINGW);
 		}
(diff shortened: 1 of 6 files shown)


ofTheo:
So this should handle a project that's project folder doesn't exist as well as one where it has a project folder with src/ folder just as well? 

Looks like an error in macOS ci might need a PR to OF/OF to be merged first?

dimitre:
Yes. I just not sure by a folder that doesn't exist. never tested this before
you mean PG should create a project folder if it doesn't exist?
about the test I think the cache is making it fail. will try to erase cache manually so it has to clone latest OF

dimitre:
@ofTheo I've pushed two important changes here. 
One, PG had an error using make because OF_ROOT was being written in config.make with double quotes., fixed
other one, it was altering OF_ROOT only if it came from template, commented out ```str.rfind("# OF_ROOT =", 0) == 0```
and I've completed with checking also uncommented also, so you get the right OF_ROOT if you once decide to move the project somewhere else and run PG again.

commandLine/src/main.cpp
```diff
@@ -244,7 +246,7 @@ void recursiveUpdate(const fs::path & path, ofTargetPlatform target) {
 	// second check if this is a folder that has src in it
 	if (isGoodProjectPath(path)) {
 		nProjectsUpdated++;
-		auto project = getTargetProject(target);
+//		auto project = getTargetProject(target);
```
ofTheo, line 249: just curious why this was commented out?

dimitre: updateProject function has the same code inside. so it was just hanging there

dimitre: I should enable unused vars for this project

ofTheo: ahh yes - expanding now I see it. Thanks!!

ofTheo:
Awesome! Can you pull in PG/Master ? 
I added PG generation of examples folder to the build script - so we can make sure these changes can still generate examples without issue.

dimitre:
Sure! let me know when changes are there and I pull in this one. no commits behind. maybe it is already up to date

PR #357, opened 2023-04-24
fixes to allow PG command line to be compiled in Visual Studio

ofTheo:
The command line app doesn't compile in Visual Studio without these change. 
Mostly issues with fs::path args being passed to string types. 

@dimitre maybe give a quick look over incase anything stands out?


dimitre:
ok, I think it should be all good for now. I've only looked but didnt pull and compile.
we can update more parts to fs::path in near future so it will hopefully support path in other languages in windows too.

ofTheo:
thanks! going to merge this to try and get Windows CI up and running

PR #363, opened 2023-04-30
[W.I.P.] more frontend refactoring

2bbb:
refactoring to find bug easily, and some bug was fixed.

### DIFF
--- a/frontend/app.js
+++ b/frontend/app.js
@@ -1,8 +1,11 @@
+// @ts-check
 // instead of ipc, maybe?
 // https://github.com/atom/electron/blob/master/docs/api/remote.md
 
+// @ts-ignore
 const ipc = window.ipc_wrapper;
 const path = ipc.path;
+const fs = ipc.fs;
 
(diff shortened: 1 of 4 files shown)


2bbb:
@ofTheo 
maybe good time to merge now.
please test on windows if you can...!!

ofTheo:
Thanks @2bbb !!
Will merge now.

ofTheo:
Windows seem to work well with these changes! 
Absolutely no issues from the projects I ran or examples updated. 

ps: Windows/VS OF seem to compile so much faster now on VS 2022 - I think maybe thanks to @dimitre's header cleanup 💪

dimitre:
Great work @2bbb! Glad Electron is finally up to date.
It would be great if sometime in the future we could port PG frontend to https://tauri.app
so we get a few megabytes app instead of 340. 
I tell this only to share this modern alternative to Electron but I myself can't do a thing in npm, electron etc.

PR #369, opened 2023-05-04
fix drag & drop

2bbb:
I found bug on drag & drop (not work now...) and fixed it.


dimitre:
Thanks for the great improvements. it is finally building the PG on apple M1.

Issue #368, opened 2023-05-02
Wrong paths for addons on XCode on latest nightly builds


dimitre:
In XCode examples, the ones who uses addons
The files are correctly linked but the folders are not, and are appearing as red.

![Screen Shot 2023-05-02 at 13 00 22](https://user-images.githubusercontent.com/58289/235720945-b4a974fb-91b2-44b0-a1cd-731b8c1ef7c2.png)

But I'm testing with the latest projectGenerator and it is including folders correctly.

dimitre:
I've just tried to test using the recursive command line like this one
```
./projectGenerator.app/Contents/MacOS/projectGenerator  --recursive -posx -o../../../../ ../../../../examples
```
and it works correctly. 
so maybe it is a Linux issue, I can look at this. 
Is the script checking out projectGenerator latest? and which command does it use to generate the examples projects?

ofTheo:
one more possibility - maybe OF master needs PG to be updated to latest?

dimitre:
Yes maybe another possibility
I've opened an Ubuntu here, pulled OF and PG to latest, compiled PG, used commandLine recursively to generate XCode and transferred here via Rsync and addon paths are OK here.

2bbb:
same problems are happen on my mac when I test external source path.

my pg frontend's verbose outputs (added some line breaks):

```
(command used: /Users/2bit/prog/of/openFrameworks/apps/projectGenerator_dev/frontend/app/projectGenerator
  -v
  -o"/Users/2bit/prog/of/openFrameworks"
  -a" "
  -p"osx"
  -s"/Users/2bit/prog/of/openFrameworks/apps/myDevApps/mySketch/src,/Users/2bit/prog/of/openFrameworks/apps/myDevApps/arcfix/src" -t"" "/Users/2bit/prog/of/openFrameworks/apps/myDevApps/myAutomaticSketch")
```

and

```
[notice ] project path is: "/Users/2bit/prog/of/openFrameworks/apps/myDevApps/myAutomaticSketch"
[notice ] setting up new project "/Users/2bit/prog/of/openFrameworks/apps/myDevApps/myAutomaticSketch"
[notice ] updating project "/Users/2bit/prog/of/openFrameworks/apps/myDevApps/myAutomaticSketch"
[verbose]  adding file /Users/2bit/prog/of/openFrameworks/apps/myDevApps/mySketch/src/ofApp.cpp in folder mySketch/src to project 
[...]

PR #371, opened 2023-05-07
WIP - Fix paths for additional sources (XCode)

dimitre:
This PR already fixes some of the issues with additional source folders on XCode.
There are some minor issues to address yet

Closes : https://github.com/openframeworks/projectGenerator/issues/372
Closes : https://github.com/openframeworks/projectGenerator/issues/368
Closes : https://github.com/openframeworks/projectGenerator/issues/366


dimitre:
Hey @ofTheo just to let you know I've been working on PG commandline issues, hopefully they will be solved in this PR but it got more complex than I initially thought so it is  huge 😅

ofTheo:
Cool! 
Does this PR close any issues? 

Also for such a large number of changes, would be good to try a few different tests. 
- Make a project outside of the OF_ROOT folder structure  
- Update a project outside of the OF_ROOT 
- Use the PG frontend to update a project with additional sources 
- Try generating the examples

dimitre:
Yes I think this closes 2 opened issues maybe some more things.
Super, those are important tests.
I finally have temporary access to a machine with vs2019, it would be great to have access to a project using "prop" files so I can compare to make it work correctly with PG

ofTheo:
@dimitre I think a lot of @elliotwoods adding use the .prop files. 
https://github.com/elliotwoods/ofxSpout 
and
https://github.com/elliotwoods/ofxCvMin 
both use them.

dimitre:
Closes https://github.com/openframeworks/projectGenerator/issues/366

dimitre:
Thanks @ofTheo, now I see where props files are located in Visual Studio 2019 (Property Manager window) so they are being added correctly, finally.
I've confirmed this PR potentially closes 3 issues.

I'm investigating other addons like ofxKinectForWindows2 and I think it is not hard to be able to add a vcxproj to the generated project, automatically with PG.
Is there some "default" path for this kind of addon files? I've noticed both ofxBullet has a folder called "ofxBulletLib" and ofxKinectForWindows2 has the folder "ofxKinectForWindows2Lib". is this the additional project default folder? addonName + "Lib" ?

2bbb:
hi @dimitre !

this is great work!

but why do you remove many `std`...?

this may not be essential, but I want to point this.
I agree to use `using namepace std` at implementation file (.cpp), but I can't agree do it on header.
so, maybe you do because those are only used in PG project, but it's not good manner I think...

I can't judge what we need to choice, this is only pointing out what I noticed.

thanks

dimitre:
Thanks @2bbb I can revert that. I've just thought this is not a problem since the PG project is a closed one, we usually don't reuse components from there in other OF projects.
Does it make sense? or maybe there are more things to consider?

dimitre:
Hello @ofTheo let's enable discussions in this repository? 
Do you prefer to keep all centered in OF one?
I have some progress in this PR and wanted to ask / share / centralize info about PG development and usage.
Thanks!

ofTheo:
good to merge @dimitre ?

dimitre:
Hey @ofTheo I had to fix some more things yesterday and today.
Now I finally think it is good to merge.
I'm planning of merging today after passing the tests, is this ok for you?

Issue #372, opened 2023-05-24
in osx, local add-ons are ignored because they don't seem to exist (but they do) 


lmtanco:
Hi there! 
I am working with the nightly v20230524_osx version of Project Generator. 
My local add-ons are correctly read by project generator out of addons.make: 

![image](https://github.com/openframeworks/projectGenerator/assets/1093084/4b92c750-d2e9-4720-b8c5-82cacc6ce7c9)

However, later they are ignored (and addons.make is actually modified  to delete all ignored addons): 

![image](https://github.com/openframeworks/projectGenerator/assets/1093084/76f992b5-660e-4973-ae56-eb477e0bb80b)

But I am sure they do exist! 

I would like to help debugging this but not sure how to build a debug version of projectGenerator on macOS. 

My configuration is: 

- of_v20230524_osx_release
- macOS Ventura 13.3.1 (M2 Pro)

Many many thanks for any help! 

Luis.

dimitre:
I'll work on this issue later but for now can you try the same thing renaming "addons" folder to another name? any other name will do, like local_addons. let me know if it works or still fail

roymacdonald:
Hi, same here. The name of the folder does not matter. This is a huge deal breaker for me as lots of my projects rely on local addons. I see that there have been a lot of changes regarding paths and etc. Not sure why this was done but the way it was before worked well. 
I can help with this. @dimitre can you please point out where to start looking at?

lmtanco:
Thanks! Just to confirm @roymacdonald, renaming "addons" did not help.

2bbb:
@dimitre
thanks to mention.
i will debug soon!

2bbb:
I tested frontend with local addons directory.

`addons.make` is

```
ofxOsc
./addons/ofxPubSubOsc
```

and results console out is:

```
(command used: /Users/2bit/prog/of/openFrameworks/apps/projectGenerator_dev/frontend/app/projectGenerator -v -o"/Users/2bit/prog/of/openFrameworks" -a"ofxOsc,./addons/ofxPubSubOsc" -p"osx"  -t"" "/Users/2bit/prog/of/openFrameworks/apps/myDevApps/pgLocalAddonsDev")
```

is this wrong behavior of frontend...? `-a` options is seems to correct I think.

roymacdonald:
@2bbb Thanks !.
does that command work? I mean does it generate the correct xcode project? (or whatever project)?
[...]

PR #373, opened 2023-05-26
path.join to path.resolve

2bbb:
maybe this works usefully when rare situation like local addon path is gived with absolute path.


2bbb:
this PR don't relate to and will not solve #372, but this problem found when I try to solve it.

PR #374, opened 2023-05-26
fix `local_addons` behavier

2bbb:
related to #372

**currently, tested on only Xcode. please test other envs...**

this PR fixes problem about `local_addons` in `addons.make`
I tested with below:

addons.make:

```
ofxOsc
./local_addons/ofxPubSubOsc
../project_common_addons/ofxUDL
```

directory structure:

```
apps/
  myDevApps/
    project_common_addons/
      ofxUDL/
        ..
    pgLocalAddonsDev/
      local_addons/
        ofxPubSubOsc/
          ..
```

and generated project:

スクリーンショット 2023-05-26 18 39 20

<s>I can't fix what `..` is appeared. but I feel like it's getting better.</s> fixed!

### DIFF
--- a/commandLine/src/addons/ofAddon.cpp
+++ b/commandLine/src/addons/ofAddon.cpp
@@ -515,7 +515,7 @@ bool ofAddon::fromFS(fs::path path, const std::string & platform){
 	// not possible today because there are string based exclusion functions
 	for (auto & s : srcFiles) {
 		fs::path folder;
-		auto srcFS = fs::path(prefixPath / fs::relative(s, containedPath));
+		auto srcFS = fs::path(prefixPath / fs::relative(s, isLocalAddon ? pathToProject : containedPath));
 		if (isLocalAddon) {
 			folder = srcFS.parent_path();
 		} else {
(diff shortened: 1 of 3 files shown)


2bbb:
https://github.com/openframeworks/projectGenerator/blob/fb2dddbedfad3ce9179323585805cbb3c34ccef7/commandLine/src/projects/baseProject.cpp#L395-L400

I added `#if` condition for envs using Xcode to fix wrong external sources directory.
but I think this removing "../" / "..\\" is not needed on other envs too.

@ofTheo @dimitre @roymacdonald 
can anyone test behavior about external source on other envs (win / linux) with remove those lines?

roymacdonald:
I don't get what is the point of removing a part of a path, but will test and report back.
Thanks for this

2bbb:
@roymacdonald 

yes, I can't understand why this code is needed.
please test!

thanks!

ofTheo:
Hmm - I think it might be due to the folders now being references in Xcode so the relative path is needed ( where as before we were creating manual groups so didn't want the path in the name ).  This is a totally wild guess . 

Happy to merge and test with the nightly as this tends to be easier.

2bbb:
maybe this problem will be fixed by #371 more generally.
thus I close this PR.

dimitre:
Thanks for the ideas @2bbb I finally resumed working on projectGenerator and I'll have (hopefully) lots of things fixed in the next few days

PR #375, opened 2023-06-05
buildPG update for subsequent running locally

dimitre:
buildPG.sh now keeps openFrameworks cloned if already exists. 
(I've uncommented this part because caching is disabled in CI now. so it will always have a fresh clone there)
and in the end it tests if the projectGenerator folders already exists, remove them if they are already there before mv part.
Again this won't happen on CI, unless we use cache between sessions.
but it can help running this buildPG locally multiple times.


dimitre:
I'm closing this one, it seems a similar feature was already implemented in another commit

PR #417, opened 2023-09-11
DLLs path fix

dimitre:
Changes in PG to address this issue
https://github.com/openframeworks/openFrameworks/issues/7652

Fixing this one will cause one issue on this another one:
https://github.com/leadedge/ofxNDI/
that can be fixed by removing this lines
```
ADDON_DLLS_TO_COPY =
ADDON_DLLS_TO_COPY += libs/NDI/export/vs/x64/Processing.NDI.Lib.x64.dll
ADDON_DLLS_TO_COPY += libs/NDI/export/vs/Win32/Processing.NDI.Lib.x86.dll
```

If we go this route I can make a PR to ofxNDI addon to ask for changes. 
thoughts @ofTheo ?

### DIFF
--- a/commandLine/src/addons/ofAddon.cpp
+++ b/commandLine/src/addons/ofAddon.cpp
@@ -149,6 +149,10 @@ void ofAddon::addReplaceString(string & variable, string value, bool addToVariab
 }
 
 void ofAddon::addReplaceStringVector(std::vector<string> & variable, string value, string prefix, bool addToVariable){
+	
+//	alert("addReplaceStringVector val=" + value + " : prefix=" + prefix, (value == prefix) ? 33 : 32);
+	if (value == prefix) return;
+
 	vector<string> values;
 	if(value.find("\"")!=string::npos){
 		values = ofSplitString(value,"\"",true,true);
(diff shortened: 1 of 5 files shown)


ofTheo:
Curious why this breaks ofxNDI ? 
Is this fix basically copying dlls automatically?

dimitre:
from what I know PG was already copying dlls automatically, but had an issue calculating DLL paths.
What is the intended default behavior?

ofTheo:
Oh I see yeah, `libs/NDI/export` is non standard for the auto copy. 
But I guess we could do both, copy dlls in the standard VS libs path and also parse dlls in the addons_make and copy them if they aren't already copied.

ofTheo:
better not to break `ADDON_DLLS_TO_COPY` as some people might have them at a global or other weird location.

dimitre:
I'll try to understand better what is happening on ofAddon::addReplaceStringVector function

ofTheo:
Hey @dimitre 

Just ran into this issue with an addon I am working on. 
Where just having a dll anywhere in my addon/libs folder would break PG. 

I'm going to merge this as I think the worst thing that could happen with the NDI example is it just copies the libs twice.

ofTheo:
Just to mention this PR fixed it for me. Thanks @dimitre

dimitre:
Yes great! In fact I'll write ofxNDI author, so we can strip paths from dylibs there

PR #418, opened 2023-09-25
Fix backslash path in MSYS2

oxillo:
use generic_string() to write paths to config.make


ofTheo:
Thanks @oxillo 
@dimitre do you see any issue merging this?

dimitre:
Thank you @oxillo 
PR seems great! let's merge it

PR #419, opened 2023-09-28
make vscode template added by default for Linux and OSX

ofTheo:
a placeholder step before making VS Code its own actual project 
makes a dent towards fixing: https://github.com/openframeworks/openFrameworks/issues/7677 

Generates VS Code by default for Linux and OS X and opens VS Code project on Linux instead of QT Creator

### DIFF
--- a/frontend/index.js
+++ b/frontend/index.js
@@ -130,7 +130,7 @@ const platforms = {
     "msys2": "Windows (msys2/mingw)",
     "ios": "iOS (Xcode)",
     "android": "Android (Android Studio)",
-    "linux64": "Linux 64-bit (qtCreator)",
+    "linux64": "Linux 64bit (VSCode)",
     "linuxarmv6l": "Linux ARMv6 (Makefiles)",
     "linuxarmv7l": "Linux ARMv7 (Makefiles)"
 };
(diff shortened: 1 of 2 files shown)


dimitre:
one more step: defining VSCode project as default in commandLine for msys2 and linux
https://github.com/openframeworks/projectGenerator/pull/422

ofTheo:
@dimitre do you know if there are differences currently between do a template vscode vs a platform vscode ? 
any advantages to keeping the template?

dimitre:
Wondering the same. I disagree with projectGenerator terminology and want to discuss this with you.
for now template is something you can solve only with certain folders copied over (from what I understand), like some shaders, some template that won't change.
and "platforms" are being used for some deep rewrite of templates.
maybe we should rename "platforms" to IDEs or something else, once it is n to n, multiple platforms can use one IDE, one platform can use multiple IDEs.

dimitre:
Now answering your answer with more attention: no advantage for keeping the template itself, since "platform" is now more complete. I'll cc vscode template author here @hiroMTB .

hiroMTB:
hello, I didn't follow recent oF development but I checked old PR and discussion regarding vscode.

here is the non merged PR and I had a discussion with Arturo and concluded using template.
https://github.com/openframeworks/projectGenerator/pull/203

hiroMTB:
and here is also another discussion and Google Spreadsheet to have a overview (os, ide, target) at that moment.

https://github.com/openframeworks/projectGenerator/pull/202#issuecomment-494106199

dimitre:
Thank you @hiroMTB I've ended up working on a similar idea of your PR without know about your efforts, about untangling ofTargetPlatform to PG platform
https://github.com/openframeworks/projectGenerator/pull/203
https://github.com/openframeworks/projectGenerator/pull/421

hiroMTB:
Thank you for your work!

ofTheo:
Going to close this in favor of a new PR based on the new work @dimitre has done with VSCode Project

Issue #420, opened 2023-09-28
Addons - Treat Linux dynamic libs .so same as .dlls 


ofTheo:
Right now you have to make a PROJECT/bin/libs/ folder for shared objects to be loaded on Linux and then manually copy them over. 

We should treat .so files that reside in an addons libs folder the same as dlls and copy them to `PROJECT/bin/libs/` 

Then Linux users can just add an addon and have it work :)

ofTheo:
closed by #426

PR #426, opened 2023-10-13
copy .so files to the project folder in linux

dimitre:
this PR tries to address this issue https://github.com/openframeworks/projectGenerator/issues/420

I've expanded the usage of getDllsRecursively to get .so files also, and moved parsing ```addon.dllsToCopy``` from exclusive template visualStudioProject to the baseProject.
it means it will try to copy .dll and .so files to the respective folders inside the project (different folders for each file extension) in different templates.

it needs testing.
if it goes well I can rename function getDllsRecursively to getLibrariesRecursively or similar.

### DIFF
--- a/commandLine/src/addons/ofAddon.cpp
+++ b/commandLine/src/addons/ofAddon.cpp
@@ -546,13 +546,16 @@ bool ofAddon::fromFS(const fs::path & path, const string & platform){
 			
 
 		}
-		if(platform == "vs" || platform == "msys2"){
+//		if(platform == "vs" || platform == "msys2"){
+		if(platform == "vs" || platform == "msys2" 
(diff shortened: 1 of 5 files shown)


ofTheo:
woohoo - this looks great @dimitre!!!

dimitre:
Yess, but untested yet @ofTheo 
do you know any addons that use .so files?
Ah I'm finally back with a windows / linux box here

ofTheo:
Ahh nice - my new addon https://github.com/design-io/ofxOrbbec has dll, so and dylibs :)

dimitre:
Yeah, it is now working but after this other PR haha
https://github.com/openframeworks/projectGenerator/pull/427

Issue #430, opened 2023-10-30
update custom icon path info in Project.xcconfig


danomatika:
The info on setting a custom icon path is out of date in Project.xcconfig: https://github.com/openframeworks/projectGenerator/blob/master/commandLine/Project.xcconfig#L29

I believe it should be:
~~~
//IF YOU WANT AN APP TO HAVE A CUSTOM ICON - PUT THEM IN YOUR DATA FOLDER AND CHANGE ICON_FILE_PATH to:
//ICON_FILE = bin/data/$(ICON_NAME)
~~~

This works fine on a project that used to set a custom icon using the original ICON_FILE_PATH variable.

dimitre:
Maybe we should remove ICON_FILE_PATH variable entirely?

dimitre:
You can create a file named App.xcconfig, it is an optional include, so settings like icon survive projectGenerator rewrites.
it is useful for app version / icon and other configurations
```xcconfig
//CLANG_CXX_LANGUAGE_STANDARD = c++20
//MACOSX_DEPLOYMENT_TARGET = 12.0

PRODUCT_BUNDLE_IDENTIFIER = org.dmtr.xp
VERSION = 0.0.7

ICON_NAME = xp.icns
ICON_FILE = $(ICON_NAME)

CODE_SIGN_IDENTITY = -;
CODE_SIGN_STYLE = Automatic;

//OF_NO_FMOD = 1
//OF_CORE_LIBS = $(LIB_GLEW) $(LIB_PUGIXML)
//OF_CORE_HEADERS = $(HEADER_OF) $(HEADER_GLEW) $(HEADER_TESS2) $(HEADER_GLFW) $(HEADER_UTF8) $(HEADER_JSON) $(HEADER_GLM) $(HEADER_PUGIXML)
//OF_CORE_FRAMEWORKS = -framework Accelerate -framework Cocoa -framework CoreFoundation -framework CoreServices -framework CoreVideo -framework IOKit -framework OpenGL -framework QuartzCore -framework Security

```

danomatika:
Yes, I do this now, however my point is that the reference in the comment is wrong and `ICON_FILE_PATH` does not appear to be used anymore, so this note should be updated.

dimitre:
Agreed! lets remove it. I was just confused because ```ICON_FILE_PATH``` is mentioned in your suggestion

danomatika:
[...]

Issue #429, opened 2023-10-30
Xcode CFBundleVersion and CFBundleShortVersionString are not equivalent


danomatika:
I've updated to oF 0.12 and notice the openFrameworks-Info.plist sets both CFBundleVersion and CFBundleShortVersionString to `VERSION`. These are *not* equivalent as CFBundleVersion is the build identifier used when uploading to App Store Connect and needs to be different every time you upload a new build for review. CFBundleShortVersionString is what the user will see as the version in Get Info, in Quick Look, etc.

In the Xcode Info.plist interface:
* Version -> CFBundleShortVersionString -> `MARKETING_VERSION` variable
* Build -> CFBundleVersion -> `CURRENT_PROJECT_VERSION` variable

I note the difference as on iOS I typically set the Version to the main version ala "1.2.0" and the Build to a simple number "1" for submitting to Apple. If I need to do another submission, either for TestFlight or because a build was rejected, I increment the Build number without having to change the actual version.

Reference: https://stackoverflow.com/a/31921249

I recommend the following:

* change CFBundleBundleShortVersionString to `MARKETING_VERSION` in https://github.com/openframeworks/projectGenerator/blob/master/commandLine/openFrameworks-Info.plist#L22
* note this difference between `CURRENT_PROJECT_VERSION` and `MARKETING_VERSION` in Project.xcconfig

danomatika:
Oops. The Build setting in the UI uses the `CURRENT_PROJECT_VERAION` variable not `VERSION`. I would propose to use`VERSION` to set the `MARKETING_VERSION` if it's found as people may be using it now. This would update what's shown in the Xcode UI and what gets built into the app Info.plist.

ofTheo:
Yes - this should be fixed. 
Happy to merge a PR with these changes.

PR #431, opened 2023-10-30
macOS: updates to version and custom app icon handling

danomatika:
This PR implements changes proposed in #429 and #430:

* separate Project.xcconfig `VERSION` into the default `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION` variables to match Xcode UI
* add examples to new method for setting custom app icons as well as info in transitioning from previous `ICON_FILE_PATH` variable

### DIFF
--- a/commandLine/Project.xcconfig
+++ b/commandLine/Project.xcconfig
@@ -18,16 +18,16 @@ DEVELOPMENT_LANGUAGE = English
 CODE_SIGN_IDENTITY = -
 INFOPLIST_FILE = openFrameworks-Info.plist
 
-// VERSIONING - uncomment to override settings in Xcode UI
+// VERSIONING - overridden if changed in Xcode UI
 
 // this is "Version" in the Xcode target Identity UI
 // suggested to use semantic versioning format ala #.#.#
-//MARKETING_VERSION = 0.1.0
+MARKETING_VERSION = 0.1.0
 
 // this is "Build" in the Xcode target Identity UI, an incremental build number
 // important for the App Store as new build submissions need a diff number even
 // if MARKETING_VERSION is the same
-//CURRENT_PRODUCT_VERSION = 1
+CURRENT_PROJECT_VERSION = 1
 
 // ICONS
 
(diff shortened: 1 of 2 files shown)


danomatika:
I was tempted to also change some formatting and add sections like:
~~~
// APP SETTINGS #####
...

// VERSIONING #####
// uncomment to override settings in Xcode UI

// this is "Version" in the Xcode target Identity UI
// suggested to use semantic versioning format ala #.#.#
//MARKETING_VERSION = 0.1.0

// this is "Build" in the Xcode target Identity UI, an incremental build number
// important for the App Store as new build submissions need a diff number even
// if MARKETING_VERSION is the same
//CURRENT_PRODUCT_VERSION = 1

// ICONS #####

// default oF app icon
....
~~~

I would also add a leading space to comments so:
~~~
//THE PATH TO THE ROOT OF OUR OF PATH RELATIVE TO THIS PROJECT.
~~~
becomes
~~~
// THE PATH TO THE ROOT OF OUR OF PATH RELATIVE TO THIS PROJECT.
~~~
... but that is a stylistic difference which doesn't have a huge impact other than my general pursuit of consistency.

ofTheo:
just fyi @danomatika 
the changes above will only affect the Project generator command line app. 

To do the same for all OF apps you would need to modify the files here: 
https://github.com/openframeworks/openFrameworks/tree/master/scripts/templates/osx

danomatika:
Why would there need to be two copies of the Xcode project files?

ofTheo:
@danomatika the ones at: https://github.com/openframeworks/openFrameworks/tree/master/scripts/templates/osx 
are what all OF projects and examples use. 

The ones in this repo are customized just for building the PG command line app and some of those settings we might not want to be global ( ie: in the same way the individual projects might want to have different Project.xconfig settings ).

danomatika:
Latest commit uncomments version variables as they act as defaults are are overridden by the Xcode UI. Also, the build variable name is corrected to `CURRENT_PROJECT_VERSION`.

ofTheo:
Awesome - thanks @danomatika !!

Issue #440, opened 2023-11-16
Problem setting ADDON_CFLAGS in Xcode projects with OF 0.12 release PG


danomatika:
I have been updating my addons for OF 0.12 and I am seeing an issue with setting `ADDON_CFLAGS` in Xcode projects for iOS and macOS and I have run into an issue with [ofxPd](https://github.com/danomatika/ofxPd).

The ofxPd [addons_config.mk](https://github.com/danomatika/ofxPd/blob/master/addon_config.mk) sets the following:
~~~
common:
    ADDON_CFLAGS = -DPD -DUSEAPI_DUMMY -DPD_INTERNAL -DHAVE_UNISTD_H -DHAVE_ALLOCA_H -DLIBPD_EXTRA
osx:
    ADDON_CFLAGS += -DHAVE_LIBDL -DHAVE_MACHINE_ENDIAN_H -D_DARWIN_C_SOURCE
ios:
    ADDON_CFLAGS += -fcommon -DHAVE_MACHINE_ENDIAN_H -D_DARWIN_C_SOURCE
~~~

After generating the Xcode project, only the last flag is set within the project's **Other C Flags**:
![Screenshot 2023-11-16 at 11 33 48 PM](https://github.com/openframeworks/projectGenerator/assets/480637/1a99e25b-046e-45cd-b9d8-079cb2c7559a)

I tried separating the flags into single lines using `ADDON_CFLAGS +=` but this does not change the behavior.

This problem basically breaks (re)generating Xcode projects which use ofxPd. For now, I have added info to the readme on [manually setting the C flags](https://github.com/danomatika/ofxPd/tree/master#of-012-and-xcode-exth-file-not-found).

danomatika:
Should I have opened this on the main openFrameworks repo? I am using the basic GUI projectGenerator included with the OF 0.12 release.

dimitre:
No, this repo is best for this issue. I'll be investigating this

danomatika:
Thanks. Nice that it was an easy fix.

dimitre:
Thanks for reporting. I'll link a note on 0.12.1 release to remember you of changing readme

PR #441, opened 2023-11-16
CFLAGS array fix

dimitre:
Address a problem where only the latest CFLAG entry was being added.
cc: @danomatika 
this is the result of ofxPd pdExample project, after the change

Screenshot 2023-11-16 at 20 04 22

Potentially closes https://github.com/openframeworks/projectGenerator/issues/440

### DIFF
--- a/commandLine/src/projects/xcodeProject.cpp
+++ b/commandLine/src/projects/xcodeProject.cpp
@@ -679,9 +679,10 @@ void xcodeProject::addLDFLAG(string ldflag, LibType libType){
 }
 
 void xcodeProject::addCFLAG(string cflag, LibType libType){
+	//alert("xcodeProject::addCFLAG " + cflag);
+	//commands.emplace_back("Add :objects:"+c+":buildSettings:OTHER_CFLAGS array");
 	for (auto & c : buildConfigs) {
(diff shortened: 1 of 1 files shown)


danomatika:
I suggest double-checking `ADDONS_CPPFLAGS` handling as well.

danomatika:
@ofTheo IMO This is needed for an OF 0.12.1 bugfix release.

dimitre:
@danomatika I've double checked everything that uses array in xcodeproject, all good.
Changes are already merged so it will be on 0.12.1
Nightly generated with changes will be available in 30 minutes
Can you please test your other addons, or addons you use frequently to see if there are any remaining PG issues? Thanks

danomatika:
I have updated all of my personal addons. I am not able to build iOS projects but I figured that's because I have an older macOS 13 and Xcode installed, so I didn't try debugging it.

dimitre:
I'm using macOS 13 too, it should be all good. let me know if you have issues with the addons, maybe there are things missing in iOS project generator

dimitre:
@danomatika https://github.com/openframeworks/openFrameworks/releases/tag/nightly

PR #442, opened 2023-11-23
removing ConfigParseState enum

dimitre:
this is one idea of removing complexity, 
now we have functions to convert back and forth ConfigParseState to string, like this:
```c++
string ofAddon::stateName(ofAddon::ConfigParseState state){
	switch(state){
	case Meta:
		return "meta";
	case Common:
		return "common";
	case Linux:
		return "linux";
	case Linux64:
		return "linux64";
	case MinGW:
		return "msys2";
	case VS:
		return "vs";
	case LinuxARMv6:
		return "linuxarmv6";
	case LinuxARMv7:
		return "linuxarmv7";
	case LinuxAArch64:
		return "linuxaarch64";
	case AndroidARMv5:
		return "android/armeabi";
	case AndroidARMv7:
		return "android/armeabi-v7a";
	case Androidx86:
		return "android/x86";
	case Emscripten:
		return "emscripten";
	case iOS:
		return "ios";
	case OSX:
		return "osx";
	case Unknown:
	default:
		return "unknown";
	}
}
```

And the functions to convert back. 
And this introduces errors, like "linuxarmv6l" turns into different "linuxarmv6" when converting back and forth and some others, 
so I decided to simplify and use only strings, having a new vector parseStates where the string can be checked if the state is valid.
Possible platforms are handled now in just one place, and I've added "tvos" so ofxSyphon doesn't return errors.


dimitre:
This conversion back and forth even has typing errors as "linuxarmv6l" will turn "linuxarmv6" which is different if converted back and forth.  "linuxarmv7l" also has this typo

dimitre:
I've now added "tvos" as valid state, since ofxSyphon is using it

dimitre:
If you have the time I would like to know your opinion on this one @artificiel

artificiel:
it seems the original code is sort-of inspired by the state-machine-pattern (albeit not really). maybe there was an intent to make things more stateful. in any case, changing the `else-if` for `std::find` is good! (std::any also, when you just need a bool "present or not" result and not an iterator). But i'm not familiar enough with PG to comment more on the design side.

however for string "defines" I find that if the strings are typed only once (like for `parseStates`) then they may be hard-coded in the vector as you did. but if the strings are to be repeated, making strings like this:

```c++
const string ADDON_NAME = "ADDON_NAME";
const string ADDON_DESCRIPTION = "ADDON_DESCRIPTION";
const string ADDON_AUTHOR = "ADDON_AUTHOR";
const string ADDON_TAGS = "ADDON_TAGS";
const string ADDON_URL = "ADDON_URL";
```
is good because it forces correctness when re-using the same string (and a typo error in a string can be hard to debug).

dimitre:
@ofTheo can you have a quick look to give your opinion? 
I think it is good to merge

ofTheo:
Looks good to me @dimitre ! 👍

Issue #449, opened 2023-12-15
VS broken in OF/OF - possibly from recent addons changes in PG


ofTheo:
So I think maybe something happened with #445 that broke the PG project generation for VS and means that the allAddonsExample is failing in OF ci. 

Looking at the commits in OF/OF both PRs which weren't merged started failing for VS and an unrelated PR for emscripten started failing for VS. 

The emptyExample is passing, but not allAddonsExample so it could be from #445 

@dimitre can you try the current PG and see if it generates a workable allAddonsExample for you on VS?

This is the output from the VS CI
https://github.com/openframeworks/openFrameworks/actions/runs/7212770989/job/19651179663

```
C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\shared\guiddef.h(161,13): error C3861: 'memcmp': identifier not found (compiling source file src\main.cpp) 
C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um\winnt.h(22421,5): error C3861: 'memset': identifier not found (compiling source file src\main.cpp) 
C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\shared\stralign.h(483,16): error C3861: '_wcsicmp': identifier not found (compiling source file src\main.cpp) 
D:\a\openFrameworks\openFrameworks\addons\ofxPoco\libs\poco\include\Poco\Bugcheck.h(73,14): error C2039: 'string': is not a member of 'std' (compiling source file src\main.cpp) 
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\include\type_traits(21): message : see declaration of 'std' (compiling source file src\main.cpp) 
D:\a\openFrameworks\openFrameworks\addons\ofxPoco\libs\poco\include\Poco\Bugcheck.h(73,21): error C3646: 'what': unknown override specifier (compiling source file src\main.cpp) 
D:\a\openFrameworks\openFrameworks\addons\ofxPoco\libs\poco\include\Poco\Bugcheck.h(73,26): error C2059: syntax error: 'const' (compiling source file src\main.cpp) 
D:\a\openFrameworks\openFrameworks\addons\ofxPoco\libs\poco\include\Poco\Bugcheck.h(73,92): error C2238: unexpected token(s) preceding ';' (compiling source file src\main.cpp) 
[...]

PR #451, opened 2023-12-15
Fix for VS broken in OF/OF

dimitre:
In progress. 
It works already, but right now it is a ugly hack
I'll work a little bit more on this one
closes https://github.com/openframeworks/projectGenerator/issues/449


ofTheo:
amazing - thanks @dimitre!

ofTheo:
Awesome - thanks @dimitre !!!

PR #464, opened 2023-12-19
ofxPoco / ofxAssimp fix

dimitre:
ok, now we can build both addons in VS
it can be merged as soon as checks are ok


ofTheo:
wow - awesome!!

PR #470, opened 2023-12-24
Open Folder button after successful Update

danoli3:
Convivence user method and experiment of deeper look into code for PG

Screenshot 2023-12-22 055335

Added as the Search folder icon linked to path is locked in context window so open folder is useful when not wanting to directly open IDE and wanting full native folder context 

Context shown is native Windows Explorer, macOS Finder or linux folder view

### DIFF
--- a/frontend/app.js
+++ b/frontend/app.js
@@ -558,7 +558,7 @@ function setup() {
 
          $("#IDEButton").on("click", () => launchInIDE());
 
-         $("#FolderButton").on("click", () => launchInIDE());
+         $("#FolderButton").on("click", () => launchFolder());
 
 
          $("#verboseOption").checkbox();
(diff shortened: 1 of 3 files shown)


frontend/app.js
```diff
@@ -1060,8 +1073,10 @@ function displayModal(message) {
 
     if (message.indexOf("Success!") > -1){
         $("#IDEButton").show();
+        $("#FolderButton").show();
     } else {
         $("#IDEButton").hide();
+        $("#FolderButton").show();
```
ofTheo, line 1079: should this be hide?

danoli3: if it fails to build, allow to direct to open folder as well maybe need to check files etc

ofTheo:
Ahh got it. 
Looks good to me then! 👍

PR #471, opened 2023-12-24
Project Generator - xcframeworks, fixed up extra platforms

danoli3:
Adding support for Apple's new xcframeworks which all the new macOS, iOS, tvOS libraries are built into 
- fixed up extra platforms (optimisation by looking at code, it will avoid going into those if target platform selected)


danoli3:
Okay cool tested this is working

PR #481, opened 2024-02-04
Fix MSBuild paths in some bash

danoli3:
Some versions of bash not collating the / path for ms/vsbuild 
Fixed by double //

- exclude .mm/storyboard vs


danoli3:
@dimitre this should fix the params were not getting parsed I think

PR #483, opened 2024-02-12
osx to macos

dimitre:
This one renames osx to macos on project generator commandline.
it will require a folder rename in OF project, 
```../../../scripts/templates/osx ```
to
```../../../scripts/templates/macos```

projectGenerator Frontend can be udpated later, as parameter "osx" will be converted to "macos" internally for now


roymacdonald:
As I commented in apothecary's issues https://github.com/openframeworks/apothecary/issues/437 having this done messes up lots of things, in particular in addons.
Having this done would mess up support for so many addons and it will become a really annoying thing for both users and addon's devs. I think that this should not be done and all should be reverted to only use `osx` rather than `macos`.

PR #484, opened 2024-03-02
XCode simplification, adding "data" folder to file browser

dimitre:
I've worked today in a simplification of adding files, folders, build phases, etc. so the code is more compact, we uniformize what we are doing.
I've added "data" folder to the file browser also, so resources like exr images are easily seen, shaders are easily edited.

### DIFF
--- a/commandLine/src/projects/baseProject.h
+++ b/commandLine/src/projects/baseProject.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#define PG_VERSION "33"
+#define PG_VERSION "34"
 
 #include "ofAddon.h"
 #include "ofFileUtils.h"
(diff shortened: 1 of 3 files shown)


dimitre:
@danoli3 reviewing the xcframework functionality I've noticed it was adding to "frameworks" vector instead of "xcframeworks"
so I changed this line:
```c++
getXCFrameworksRecursively(libsPath, frameworks, platform);
```
to this
```c++
getXCFrameworksRecursively(libsPath, xcframeworks, platform);
```

danoli3 (approved):
Looks good!!

dimitre:
Nice, thanks! there is a tiny regression I have to fix on this one, 
it is not checking "code sign on copy" but I can fix in the next PR.

PR #503, opened 2024-05-28
XCode addon folder fixes

dimitre:
This fixes an issue with PG where XCode addon folders full path are not preserved.
now each addon will be relative to the group ("addons folder") that by itself is related to the project.

now the arrow will point to the correct folder, and we will be able to use right click -> show in finder in any addon folder.
There will be another PR in OF repo fixing the template

Screenshot 2024-05-27 at 22 48 46

### DIFF
--- a/commandLine/src/projects/baseProject.h
+++ b/commandLine/src/projects/baseProject.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#define PG_VERSION "38"
+#define PG_VERSION "39"
 
 #include "ofAddon.h"
 #include "ofFileUtils.h"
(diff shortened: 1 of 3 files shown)


dimitre:
@danoli3 this fixes some issues with frameworks and xcframeworks link phases

armadillu:
@dimitre  Hey, I am unsure which commit this happened in, but I cant seem to add addons that are not in the "OF/addons" folder (ie "local_addons") any more. I have a bunch of automated scripts that create and compile example projects for all my addons, (which I keep in a different directory, outside the OF repo) and the latest PG fails to include them correctly now. This was working ok on PG version "34".

I have the PG xcode project setup so I can debug the creation of an example project (with custom scheme launch arguments), but man it is messy in there. 

image

I am happy to help but I wanted to check where are you with this (ie are you mid-update?)

I like the idea of making all paths relative to their "group" folders, but the problem I am seeing is that the "local_addons" main folder that is created doesn't see, to be set correctly. If I manually set it in Xcode after the project is created, it seems to work ok.

dimitre:
Hey @armadillu I'll be investigating now. Thanks for reporting

dimitre:
Nice, now the folder immediatly after local_addons or external_sources, or root have an absolute path. 
The rest of them are relative to group. I'll be merging soon today and it will be v.43

dimitre:
It should be working OK now @armadillu

armadillu:
Great, seems to work on my end!

PR #512, opened 2024-05-30
WIP Ubuntu 24.04 actions fix

dimitre:


danoli3:
24.04 now building on master

Issue #521, opened 2024-06-08
Clean on Xcode not working due to bin/data lacking attributes 


danoli3:
### Could not delete `bin` because it was not created by the build system and it is not a subfolder of derived data.

<img width="1202" alt="Screenshot 2024-06-08 at 5 19 07 pm" src="https://github.com/openframeworks/projectGenerator/assets/830748/c20e07a1-f89a-4b26-bae5-3012b6897239">

```

Showing All Messages
Could not delete `openFrameworksmacOS/apps/myApps/emptyExample/bin` because it was not created by the build system and it is not a subfolder of derived data.

```

Fix is applying an xattrib to the folder in the template and copying with cp -X  or rsync.

I might just add to the projectGenerator code too with system command just to make it easier and bound in c++ rather than folder attributes which might not be pushed to git or zipped not sure

danoli3:
std::string command = "xattr -w com.apple.xcode.CreatedByBuildSystem true " + templateBinDir.string();

danoli3:
One could also add to the Xcode build phases like:
```
mkdir -p "${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/bin"
xattr -w com.apple.xcode.CreatedByBuildSystem true "${TARGET_BUILD_DIR}/${UNLOCALIZED_RESOURCES_FOLDER_PATH}/bin"
```

PR #520, opened 2024-06-07
v46 xcframeworks 

dimitre:

PR #522, opened 2024-06-08
V46 XCFrameworks Fixing 

danoli3:
Adding to @dimitre's https://github.com/openframeworks/projectGenerator/pull/520

This fixes Addon embedding for xcframeworks for iOS/macOS/tvOS 

Also fixes https://github.com/openframeworks/projectGenerator/issues/521


dimitre:
just one observation here
```
if (fileType == "wrapper.framework" || fileType == ".xcframework") {
```
should be
```
if (fileType == "wrapper.framework" || fileType == "wrapper.xcframework") {
```

PR #536, opened 2024-07-03
projectGenerator linux fixes

danoli3:


danoli3:
Fixes issues on linux / macOS changes for libs download xcframeworks

PR #540, opened 2024-07-10
project override for template objects like plist fixes - compare and …

danoli3:
project override for template objects like plist fixes - compare now

manually add keys if missing from deployed file vs template (if we add new ones to template, it should add them) - not change the key value though, if does exist 

Fixes: https://github.com/openframeworks/projectGenerator/issues/542 with auto oF Path finder

### DIFF
--- a/commandLine/src/main.cpp
+++ b/commandLine/src/main.cpp
@@ -165,14 +165,36 @@ bool isGoodProjectPath(fs::path path) {
 
 bool isGoodOFPath(const fs::path & path) {
 	if (!fs::is_directory(path)) {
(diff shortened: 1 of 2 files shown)


danoli3:
Still testing

danoli3:
Okay tested and projectGenerator - plist formatting is kept, it doesn't update variables if different in target file now. 

Added automatic openFrameworks folder find code if projectGenerator is placed within any subfolder of openFrameworks - Solve issues with oF path

danoli3:
Fixes : https://github.com/openframeworks/projectGenerator/issues/542

PR #548, opened 2024-07-13
v60 - Fix codesign issues Mac

danoli3:
- Disabled fmod for causing linking crashes when running in isoation
- build bash scripts updated to include rsync copying function to include codesigning attributes
- commandLine project running copy of commandLine to every build rather than incremental that was missing code signed version and subsequently not updating
- ci build bash scripts to use rsync to retain code signing attributes macOS 
- chmod +755 on projectGenerator commandLine


danoli3:
further analysis of VS pathing issue:
```
[notice ] ofPath auto-found and valid using [/Users/one/SOURCE/openFrameworksmacOS]
[notice ] OFRoot set: ["../../../.."].
```

Look like we go from a normalised path to relative and might be the cause 
cause it gets weird when dont pass the -o
[notice ] OFRoot set: ["/Users/one/SOURCE/openFrameworksmacOS/apps/projectGenerator/commandLine/bin"].

danoli3:
verify codesign for moved commandLine 
```
app/projectGenerator: invalid Info.plist (plist or signature have been modified)
In architecture: arm64
```
re codesign
```
app/projectGenerator: replacing existing signature
app/projectGenerator: signed Mach-O universal (x86_64 arm64) [projectGenerator-55554944407b18a4ee293ce9ace0a21d13094990]
app/projectGenerator: valid on disk
app/projectGenerator: satisfies its Designated Requirement
Code signing completed and application moved to *
```

So thats for valid signed. 
However I've noticed the 5.8mb is randomly not signed at all either:
```
./projectGenerator: code object is not signed at all
In architecture: arm64
./projectGenerator: signed Mach-O universal (x86_64 arm64) [projectGenerator-55554944407b18a4ee293ce9ace0a21d13094990]
verify codesign
./projectGenerator: valid on disk
./projectGenerator: satisfies its Designated Requirement
```

danoli3:
Added correct entitlements for use of: Desktop/Downloads/Documents
Added Full disk access but thats up to user to allow via settings if warranted. Should solve most of the permission problems running from downloads etc

PR #551, opened 2024-07-14
VS Project dir path out of directory 

danoli3:
Fix for addons?


danoli3:
Seems like it fixed the out of folder bug! merging

PR #559, opened 2024-08-01
projectGenerator 67 - UI fixes / Function Fixes and CI fix

danoli3:
PG 67. 
- Fix for Default platform. 
- oF path now calling to CMD to determine auto path at start
- PG Version shown in settings
- other Variables shown in settings.

image

When in update mode. Show folder/open ide button now visible always. 
- Fixes for code exceptions.

Just noticed CI is failing so trying to fix that... always have so many commits I haven't pushed arrrrr


danoli3:
Got some time now, finished with emscripten game project jam yay!

danoli3:
Also tooltips and console now work again

danoli3:
Also in terms of use case for the future of this Chrome powered ProjectGenerator...

I was doing some profiling and it's the same as a TAB of Chrome open.. that is appealing 

And has the potential to run... emscripten directly into a tab of it... call shell... call bash... call other programs like ofAddons / ofxPackageManager and get all the functionality of those if addon detected or embedded 

Hell we could even embed the examples and make call in this for emscripten or Make or cmake. I can see the path

And this with a bit of re-mixing we can make this good !

dimitre:
I think projectGenerator in nightlies (v63) can't create an empty folder for a new project, but it seems to be fixed in v66
so maybe it is just a matter of bundling latest for fixing PG in nightlies?

danoli3:
Yeah the ci wasn’t building till today

PR #569, opened 2024-08-29
OFProjectGeneratorVersion back in XCode project

dimitre:
we now have back in project.pbxproj a way of knowing which version of projectGenerator was used to build.
It will be useful to debug problems in projects from nightly builds
```json
{
	"_OFProjectGeneratorVersion": "0.73.0",
	"archiveVersion": "1",
	"classes": {},
	"objectVersion": "54",
}
```


danoli3:
apothecary vs release x64 libs seem to have not deployed, need to replace that release action, rebuilding and ill trigger action when its done

Issue #577, opened 2024-09-24
Project generator has become unreliable


roymacdonald:
Hi guys,
first let me say that I truly and deeply appreciate all of your efforts to make OF's ecosystem a better one, in particular @dimitre and @danoli3 efforts.

Although, PG has become unreliable to me.
It worked perfectly well for so long, and while I see that many things need to be updated in order to make it work with newer OSes and IDEs many things broke in the process that did not need to get broken in the first place. Even when I use OF by pulling from github quite frequently, I still need to rely on a quite old version of PG. 
For me the thing that most bothers me is that local addons no longer work. And at some point I noticed that the parsing addon_config.mk was not done properly and it missed some of the things in it (although I see that got fixed now).
Now, I have found that Apothecary is naming what used to be */libs/osx to */libs/macos. I think that that is a really bad move, as I already expressed [here](https://github.com/openframeworks/apothecary/issues/437). A lot gets broken, for something that looks like an aesthetic change mostly (is there really a  technical need for such rename?)

I have already spent several days digging into PG's code and I have to say I am not pleased with what I have found. I see a lot of patched code, some big changes that broke lots (like switching from using strings to use std::filesystem), lots of super redundant code, big chunks of code that are copy/paste,  lack of encapsulation, lack of a proper test bed, and the list goes on.

I see that there are some testing scripts but they only test if PG can generate a project for each example but it is not testing if these where properly generated.  and still, the examples dont cover the whole use cases (there are many addon's features not being used by the examples)
How can we properly test it works? (this would require not only bulding PG but also generating projects that  test all the features of PG and have this to compile AND run).

In general I see that since @arturoc left OF's dev lead a lot of its development became messier. 

And please don't get me wrong, I am not trying to blame anyone, I am quite sure that all of you do what you do with the best intentions, but we need to be a bit more careful.

What are your thoughts about this @ofTheo @danoli3 @dimitre  ?

dimitre:
I think OF is unreliable today in this state. Libs are not working in the latest months.

PG is not unreliable for me, today. 
There were this recent issues difficulties with nlohmann (recently fixed)
I know there is the ../ thing you mentioned 
and I'm glad you are now working on some PR code, so you too can contribute to the closing of this issue here.

Feel free to open one issue for each detail you know it is not working 100% on PG and I can help fixing or discussing what can be done. 

In my opinion we can work in the following sequence:
Once the libraries are in a stable position (they are not today)
[...]

PR #578, opened 2024-09-25
Addons parsing improvement (dont merge yet)

roymacdonald:
This is what I have done so far related to [this](https://github.com/openframeworks/projectGenerator/issues/577)

I have encapsulated the ofAddon class so it is easier to test that it is loading/parsing properly the values for any addon. 

I have dealt partially with the macos/osx issue. I say partially because I have not tested it thoroughly and it mostly feels like a hack. But so far it works for me, I am able to use addons, local addons, addons that use libraries, xcframeworks. Still neet to test for those using frameworks, system frameworks and dylibs. Currently only testing on macos 14 (sonoma) and xcode 15.

I will setup my windows/linux computer to be able to test on those other environments as well.

The main idea behind all of it is that an addon is an addon regardless of where it sits in the filesystem. Since all the routes are relative it should not matter where these sit, thus the local addons are treated in the exact same way as the non local ones. 

Then the baseProject class does not do much with these besides loading and then it is the job of each baseProject-derived-class to override the necessary functions so it adds the addons properly. 

There is a lot to be cleaned and reduced since there is a lot of duplicated code and stuff not even used.
Eventually it should provide a much cleaner way of dealing with addons and project creating in general.


commandLine/src/projects/CBLinuxProject.h
dimitre, line 1: One suggestion: 
It is great to use override on children but not virtual, we are not using a project based in other project, all of them derive from the same base class. So it can be final, or just without virtual

roymacdonald: Hi @dimitre , true, the `virtual` is not needed, but I prefer to use it as it is clear that it is overriding another function. as far as I know it is a good practice to put it.

dimitre: in fact you use virtual in the class that will be overriden (the base class) only
the derived class will only be virtual if another class will be derived from them, which is misleading in this specific case

dimitre: the "override" word makes it clear it is overriding the base class.

roymacdonald: I had to double check this idea and found out that I am wrong, according to the C++ guidelines, although other sources said it was ok having virtual and override together. Well not completely wrong, it is just redundant to use virtual and override in the same function, and does no harm. I guess it comes from explicitly declaring as virtual an overrider function from the days when there was no override keyword. Thanks for pointing out.

ofTheo:
@roymacdonald has this been tested more since the PR? 

For testing Frameworks: 
https://github.com/astellato/ofxSyphon 

For testing dylibs:
https://github.com/design-io/ofxOrbbec

I'd love to merge this but would be good to know how much it's been tested across platforms ( as there are so many changes it's hard to review ).

roymacdonald:
@ofTheo 
I am using those addons, as well as some others I have arbitrarily chosen to check if all is added correctly. 
As said elsewhere, I need to test for other platforms, I will try to setup that during the weekend, but I feel the need to establish some sort of "test suite", like a list of the addons that would be good to test on each platform which use the different features, etc.
And having a script for automating such too. 
I will try to write that too.
So far, 
- ofxOpenCv (for static libs) (all platforms)
- ofxSyphon (for local frameworks being added) (macos)
- ofxHapPlayer (for dylibs and dlls) (macos, linux, ??)
- ofxNDI (dlybs, flags) (macos, ...)
- ofxGStreamer (for addon sources and flags) (macos, linux)
- ofxAudioUnit (for adding system frameworks) (macos)
- ofxPoco (a lot of flags and setup for all platforms) (all platforms)
- ofxOrbbec (dylibs) (platforms?)

Any other recomendations for win/linux/ etc?

dimitre:
I've now tested the PR and it is working fine with relative addons and additional source folders
I'll left some other suggestions, mostly removing commented out code so we can merge.

commandLine/src/addons/ofAddon.cpp
```diff
@@ -772,65 +801,148 @@ void ofAddon::parseLibsPath(const fs::path & libsPath, const fs::path & parentFo
...
+//			folder = fs::path { "local_addons" } / fs::relative(fFS.parent_path(), parentFolder);
+//		}
+//
+//		filesToFolders[f] = folder;
+        addToFolder(f, parentFolder);
 	}
 }
+string ofAddon::cleanName(const string& name){
+    auto addonName = name;
+#ifdef TARGET_WIN32
```
dimitre, line 833: this define is wrong, so if not addressed now at least keep the comment
// FIXME : not target, yes platform.
needed if somebody is generating windows projects in a linux/macos machine

roymacdonald: I just took that from what was there before.

roymacdonald:
Thanks @dimitre for reviewing. On what platform are you testing?
I was able to get my windows machine setup yesterday so I will be testing during the week on Windows.

ofTheo:
This would be great to merge if ready. 
Currently PG is broken on macOS with any addons that have libs  ( might be broken on other platforms too )

roymacdonald:
I've noticed a few more things that need to be addressed, although it probably is. better than what currently there is 
I will try to get these things fixed and tested throughout the week

commandLine/src/addons/ofAddon.cpp
dimitre: yes, just keep the comment and it can be addressed soon

ofTheo:
@roymacdonald let me know when its good to merge this. 

Do you know if this also addresses the issue of make not working on macOS / osx? ( see https://github.com/openframeworks/openFrameworks/issues/8155 )

roymacdonald:
Hi @ofTheo, I have been using it a lot both on windows and macos and works fine for a lot of the things that had issues, but not linking addons dylibs correctly. I am working on it now. 
As for that make issue you mention, seems related to the macos/osx folder renaming (which I still find it to be nonsense, TBH). I've tried make on macos and works fine,

ofTheo:
Thanks @roymacdonald ! 
Yeah agree the macOS / osx thing should have been handled better and done in one go with a lot of testing and 100% backwards compatibility with osx. Now it's a bit of a mess.

roymacdonald:
Indeed it is a mess.  In terms of PG, the macos/osx folder renaming is handled for most cases, but still it makes a mess in the scripts and templates folders, with lost of duplicates, and unclear which one is used, etc.

roymacdonald:
So far I've tested:

- Macos(xcode)
    - [x] addons
    - [x] local addons (arbitrary relative path)
    - [x] static libs (xcframeworks and .a)
    - [x] dynamic libs(linked and copied into app bundle)
    - [x] additional sources

- Macos(make)
    - [x] addons
    - [x] local addons (arbitrary relative path)
    - [x] static libs (xcframeworks and .a)
    - [ ] dynamic libs(linked and copied into app bundle)
        this step needs to happen as part of the make build config we have, and make an upgrade to it rather than injecting that code into it via PG. But this is how it has been all the time.
    - [x] additional sources

- Macos(VSCode)
    relies on make. So what generates the projects is fine.

- iOS, tvOS, (other apple devices)
    Nothing tested on it. I don't have a device for testing


- Windows (VS)
    - [x] addons
    - [x] local addons (arbitrary relative path)
    - [x] static libs
    - [x] dynamic libs(dll's linked and copied next to executable)
    - [x] additional sources

- Windows (make)
    not tested so far. BEcause of how it works, PG is not changing anything in there, just copying the templates. Thus, independend of PG

- WIndows(VSCode)
    Not tested so far, although uses make.

- Linux(QT)
    Nothing tested on it.

- Raspberry PI
    Nothing tested on it, yet relies on make.

- Android
    Nothing tested on it.


@ofTheo  @dimitre @artificiel

dimitre:
Nice. I think we should have it merged before resolving dylibs.
the problem with dylibs is most of the time they use some built in paths. like in the case of @rpath/ffmpeg/lib/osx/libavutil.56.dylib in ofxHapPlayer

maybe it needs more discussion. my opinion is we should remove all paths from dylibs and use them in a specific copy phase, I don't remember if frameworks or executables. it is easier to correct paths this way and dylibs find each other correctly if we use this way.

roymacdonald:
The "@rpath" issue can be handled by the addon_config.mk via LD_FLAGS. Actually, I've been testing with ofxHapPlayer and it compiles and runs without any extra setting. In that particular case, since it sets the @rpath to the dylibs that are in the addon's folder it works with make command too. 

I think that dealing with the libs paths should be handled by each developer that wants to use a dlyb in their addon/app, and we should provide instructions on how to achieve such easily.

artificiel:
this is a nice test: https://forum.openframeworks.cc/t/addon-makefile-clash/44076

ofTheo:
Yeah definitely fine to merge @roymacdonald - especially as we are not regressing with the make file dylib stuff. 
Really excited for this one!

roymacdonald:
> this is a nice test: https://forum.openframeworks.cc/t/addon-makefile-clash/44076

I was actually testing this 2 days ago. it only afects make for some reason and not xcode.  It requires a small change in the makefile to make it work.  I will add a PR

roymacdonald:
I will merge this. It seems that there are a lot of current issues that are solved here. Even when I have not tested it all I am confident that the areas taht I did not are the ones that I did not change.

roymacdonald:
damn, I don't have write access here. @ofTheo @dimitre , can you merge please?

commandLine/src/projects/CBLinuxProject.h
```diff
@@ -13,9 +13,9 @@ class CBLinuxProject: public CBWinProject {
 public:
 	CBLinuxProject(const std::string & target) : CBWinProject(target) {};
 
-	bool createProjectFile();
-	void addInclude(const fs::path & includeName){};
-	void addLibrary(const LibraryBinary & lib){};
+	virtual bool createProjectFile() override;
+	virtual void addInclude(const fs::path & includeName) override{};
+	virtual void addLibrary(const LibraryBinary & lib)override{};
```
dimitre, line 18: we can remove virtual here once it is final

roymacdonald: I just updatetd it. THanks for pointing out. Although, the linux test fails immediately when trying to pull ubuntu

dimitre: Thank you @roymacdonald let's merge it.

dimitre:
I've noticed one error here compiling with make:
```
duplicate symbol 'toString(LibraryBinary const&)' in:
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/android2024.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/CBWinProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/CBLinuxProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/main.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/baseProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/utils/Utils.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/VSCodeProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/addons/ofAddon.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/androidStudioProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/visualStudioProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/xcodeProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/qtcreatorproject.o
duplicate symbol 'toString(std::__1::__fs::filesystem::path const&)' in:
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/android2024.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/CBWinProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/CBLinuxProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/main.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/baseProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/utils/Utils.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/VSCodeProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/addons/ofAddon.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/androidStudioProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/visualStudioProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/xcodeProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/qtcreatorproject.o
duplicate symbol 'toString(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&)' in:
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/android2024.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/CBWinProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/CBLinuxProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/main.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/baseProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/utils/Utils.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/VSCodeProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/addons/ofAddon.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/androidStudioProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/visualStudioProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/xcodeProject.o
    /Volumes/tool/ofw/apps/pgd/commandLine/obj/macos/Release/src/projects/qtcreatorproject.o
ld: 3 duplicate symbols

```

it seems to work ok if we change toString functions and removeDuplicates to static
I can to a PR with this and move this functions to inside ofAddon, because it is the only place that uses it

roymacdonald:
Nice catch. I was not using make for building PG, nor tested it.
Making those static would be a good idea, but also putting those toString functions inside a namespace, something like of::priv:: which is used elsewhere.
If you can make those changes it would be great

dimitre:
ok merged in another PR here https://github.com/openframeworks/projectGenerator/pull/583
I'm manually generating a nightly build so we can inspect examples to see if they are OK in windows / mac

roymacdonald:
awesome. Thanks @dimitre

dimitre:
Thank you for the improvements in PG. 
when you have time take a look at the issues to close the ones addressed in your PR

Issue #515, opened 2024-06-05
ProjectGenerator add-on libs  pathing issue and src subfolders not linked correctly 


danoli3:
For Addon libs:
<img width="1245" alt="Screenshot 2024-06-05 at 2 08 09 am" src="https://github.com/openframeworks/projectGenerator/assets/830748/9eed664d-6dd1-4563-b43c-55027032f87b">

Where it should look like this:
<img width="800" alt="Screenshot 2024-06-05 at 3 10 35 am" src="https://github.com/openframeworks/projectGenerator/assets/830748/a3d82e6e-b869-4dd2-983e-f995ab98054b">

had a Quick Look looks related to the paths not being correct 

```
                    if (folder.begin()->string() == "addons") {
                        addCommand("Add :objects:"+thisUUID+":sourceTree string <group>");
                        fs::path addonFolder { fs::path(fullPath).filename() };
                        addCommand("Add :objects:"+thisUUID+":path string " + ofPathToString(addonFolder));
                        // alert ("group " + folder.string() + " : " + base.string() + " : " + addonFolder.string(), 32);
                    } else {
                        addCommand("Add :objects:"+thisUUID+":sourceTree string SOURCE_ROOT");
                    }


```

I re-added this to xcframeworks and now it works 

Also if you run the projectGenerator against the projectGenerator (inception)
It fails to link or add the other src folders in the same directory 

<img width="336" alt="Screenshot 2024-06-05 at 11 22 45 am" src="https://github.com/openframeworks/projectGenerator/assets/830748/eef55064-50cc-4957-bdec-baf8722e8723">

danoli3:
another issue if ofApp.cpp is modified to ofApp.mm or main.cpp is modified to main.mm it doesn't update ref to those objective-c++

danoli3:
seems like issue happening too with frameworks.... maybe some issues after code re-write @dimitre XCode more changes : PG v40 (#504)

This is just for addons

danoli3:
Do you think this is a UUID issue?

What about this works vs the prior code:
```
addCommand("Add :objects:"+thisUUID+":sourceTree string <group>");
                        fs::path addonFolder { fs::path(fullPath).filename() };
                        addCommand("Add :objects:"+thisUUID+":path string " + ofPathToString(addonFolder));
                        ```
                        
                        something to do with group maybe

dimitre:
I'll be testing this tomorrow.
[...]

PR #586, opened 2024-11-04
V82 - Simplify Local SDK Frameworks handling in osx

dimitre:
this PR simplifies local SDK Frameworks handling, point them to be relative to the correct SDK Folder set by the project.
it merges the almost similar functions to add .framework and .xcframework to be parsed in the same recursive iteration in all folders.

We can now pass system frameworks (SDK) as a parameter for a specific project in PG, so we can experiment per project 
-f"Vision,MetalKit,Metal"


danoli3:
Looks good, will this also sort not adding the XCFrameworks to the project view. Just adding as XCFRAMEWORKS search path not doing it

danoli3:
https://github.com/openframeworks/projectGenerator/issues/515#issuecomment-2155229784

PR #592, opened 2024-11-12
properly fixing vscode project. plus some extra comments.

roymacdonald:
correctly extending baseproject and adding addons to the VSCode workspace

### DIFF
--- a/commandLine/src/projects/VSCodeProject.cpp
+++ b/commandLine/src/projects/VSCodeProject.cpp
@@ -131,9 +131,8 @@ bool VSCodeProject::loadProjectFile(){
 	cppProperties.load();
 	return true;
 }
-
-
-void VSCodeProject::addAddon(ofAddon & addon) {
+ 
+void VSCodeProject::addAddonBegin(const ofAddon& addon) {
 //	alert("VSCodeProject::addAddon() " + addon.name, 35);
 
 	workspace.addPath(addon.addonPath);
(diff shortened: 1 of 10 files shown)


roymacdonald:
TL;DR

There is a lot of logic and parsing in addAddon that needs to be done in order to add get all the projects working. when it was virtual there was a lot of duplicated code in each of the different projects etc. So, the idea is that what gets extended by each project class are [these](https://github.com/openframeworks/projectGenerator/blob/master/commandLine/src/projects/baseProject.h#L146C1-L155C11) . Then also, only when something different to what is implemented in [these functions of baseProject,](https://github.com/openframeworks/projectGenerator/blob/master/commandLine/src/projects/baseProject.h#L128C1-L146C1) those should be overriden.

In the particular case of VScode project, I override addAddonBegin, which is there mostly because it was needed in VisualStudioProject to have it happening at the begining of the addAddon function. that is why it is named like that. In the case of VSCodeProject I choose to use it just because it has no clear intention as opposed to something like addAddonFramework, thus serves the purpose. As there is no addAddonToWorkspace or something like that, which feels unnecesary to add and bloats the baseProject class.

In any case, the idea is to explicitly not override addAddon function.

the addAddon functions are still public simply because these are being called from main.cpp, which is a mess I did not got all the way through but I have a good idea of what is going on (there is a lot of duplicated code, and a lot of code that never gets accessed.

@ofTheo @dimitre please merge this

roymacdonald:
@ofTheo can you please give me access to this repo ? so I can merge the cleanups, documentation, etc that is needed here without having to bother you

ofTheo:
Done! :)

roymacdonald:
@ofTheo thanks!

PR #594, opened 2024-11-19
V86 - fix adding xcframeworks to projects

dimitre:


dimitre:
addAddonFrameworks was not adding xcframeworks, only frameworks
it was missing after I started using one function to recursively add both.

Issue #580, opened 2024-10-17
macOS: local addons ignored (nightly 20241016) 


lmtanco:
Hi there, 

This is really exactly the same issue as #372, but since @dimitre asked to open a new one in case I had problems with addons again, here I am :) 

When importing the project with the projectGenerator (nightly 20241016) the local addons are detected. There is no "../" in path to the addons, so not sure how this relates to #562: 

![screenshot 1 - importing](https://github.com/user-attachments/assets/c09b8576-e2f6-48de-8b9f-5eaf75a13772)

However in the importing process it local addons are ignored and deleted from addons.make: 

![screenshot 2 - after import](https://github.com/user-attachments/assets/4dd68291-f57d-4be6-9fe8-2a23a9ce8959)

Many thanks!

Luis.

dimitre:
Great thanks. if you can please post all the command line invocation (all this text in "Success!" window textarea) I can investigate further if it is a question of CLI or frontend

roymacdonald:
Hi @lmtanco, This is a known issue.[ I've been reworking PG ](https://github.com/openframeworks/projectGenerator/pull/578)and fixed this issue among other, but still have not merged it. 

As a quick and dirty hack for it to work. Try renaming that addons folder in your project to local_addons and change it too in the addons.make file

prisonerjohn:
Hello! FWIW I'm seeing the same issue on Windows today. I see that the referenced PG has been merged, but I don't think it fixed this issue.

dimitre:
Maybe working now

PR #595, opened 2024-11-26
Local addons ok

dimitre:
I'm now working on this issue https://github.com/openframeworks/projectGenerator/issues/580
I've noticed it is mostly working after @roymacdonald PR, 
but it enforces using ./ in the start of the path, like
```
./addons/ofxMidi
etc
```

this PR allows addons to be added like 
```
addons/ofxMidi
addons/ofxTetris
```

and also removes the function cleanName.
This function actually parses a commented out line like ```#ofxSyphon```, removes the comment and add the addon back which contradicts the person original intention.


roymacdonald:
> and also removes the function cleanName.
> This function actually parses a commented out line like `#ofxSyphon`, removes the comment and add the addon back which contradicts the person original intention.

The purpose of this is to be able to have extra information of an addon in the form of a comment and not loosing that. For example, there is a very nice "Package Manager" made by @thomasgeissl which relies on this feature, where it encodes as a comment the version or commit of an addon, so you can then have that. I am not sure if I implemented it all but the idea is that you dont remove the comments from the addons.mk file

dimitre:
cc: @danoli3 @ofTheo I think this issue about a commented out addon being added should be addressed

dimitre:
@danoli3  can you help us checking why macos runner is failing here?

artificiel:
going over this in the context of #8313 

I am new to ProjectGenerator but @dimitre I think you misinterpret the `cleanName()`: it is not to process commented-out addons like `#ofxSyphon` but to trim lines with comments like `ofxSpyhon # tested as of 14.4`, where you want to process the addon as `ofxSyphon`, but you want to carry the whole line and preserve the associated comment.

danoli3:
Okay lets fix the clean name thing later just add a command to toggle that on or off passed to PG easy with default being on as before but lets merge this I'll fix the rest keep it moving

PR #605, opened 2025-03-02
iOS targets fix

danoli3:
I think this should allow platform iOS to function for PG 

Aside from the other iOS project stuff:

@dimitre saw you just did storyboard changes yay! I can add to this for iOS fixes with cherry pick straight away if you wanted to merge this / that and continue with other fixes


dimitre:
It is best to see all other fixes first. I'm keeping everything in one PR so we don't have to rebase

danoli3:
Also tested auto tag builds before any new PG updates, lets keep these ones for now: https://github.com/openframeworks/projectGenerator/releases/tag/v0.86.0

! just there is a macOS final target issue currently: Exit code 127 - command not found weird

danoli3:
I'll leave the rest to you! I just saw in my mind it was that platform == osx thing in addons

danoli3:
addon.prepareForWrite(); should get rid of these doubles its adding as well 
cherry pick this onto ur PR 
0306bdafd424865d99cb8e6fd5b9951479827bf7

dimitre:
what it is trying to fix? how to duplicate your particular issue? 
Can you try with the PR ? I've seen something about duplication before .storyboard change but it is now ok

danoli3:
oh it just forces a call on clean up dupes before write out for addon into project

dimitre:
ok I thought it was a fix by the name iOS targets fix

PR #606, opened 2025-03-03
Xcode / iOS Fixes v87

danoli3:
# ProjectGenerator 
- restored disputed clean name functionality - added command -"n" --cleanname to disable 
- Test osx command line fixes (error 127 on Actions)
- Xcode fixes for copy/add for known and used types. (.modulemap. .bundle, .tdb, .swift, .plist)
- Swift added set flagged  with c++ swift interop 5.9[ (C++ can be used)](https://www.swift.org/documentation/cxx-interop/)
- Added auto compile flags for non-arc terms (search for non-arc commands in .mm.cpp.m)
- Added auto objective-c++ flag for .cpp if obj-c commands
- Added function for checking file for objc++
- VS fixed up exclusions for Apple types for projects / cross 
- xcode kept default type for addSrc flag to disable add phases - test
- added other apple platforms


dimitre:
Please bump version for each merged PR so we can track better if any issue arises

PR #609, opened 2025-03-04
Fix PG macOS 

danoli3:
- Anti gremlins 
- remove objectivec++ -x till figure out


danoli3:
Sweet! release grelmins be gone

PR #613, opened 2025-03-04
XCFrameworks added ok in iOS and macos (v93)

dimitre:
closes https://github.com/openframeworks/projectGenerator/issues/603
tested with assimp both in osx and ios

added compile_commands.txt so ClangD is being helpful on ZED (you can install in VSCode)

More details:
.xcframeworks were not used before on macos / ios projects
now both platforms macos / ios uses the same xcframeworks, but their "templates" are distinct files with different UUIDs for each phase. so we had to differentiate them.

The important change here begins with ```if (fp.linkBinaryWithLibraries)```
so this ```fileProperty``` will add all the important stuff that needs to be added, considering each platform, and not anymore by file extension ```.framework / .xcframework```


dimitre:
@danoli3 it is fixed
and I reverted https://github.com/openframeworks/projectGenerator/pull/611

danoli3:
wait so UUID is unique per template project then should be stored for each project should have 
- osx template UUID
- iOS template UUID
- tvOS template UUID
- macOS template UUID right ?

else it wont find these or did u make it dynamic

dimitre:
I don't know nothing about tvOS. is this using the same template as iOS? if yes it uses the same UUIDs.
at least is seems (visually) to have originated from the same original project.
macos and osx I really don't understand the differences and why there are two now. 
it would be great to know from you and what are the ideas for the release. I suppose they are the same

PR #624, opened 2025-03-12
macOS build action fix

danoli3:
# PG 0.95.0
- fixing scripts for macOS deployment 
- fixing builds for Mac-arm64 seperate to universal gui 
- adding builds for linux-arm64 / linux-armv7l - gui


danoli3:
macOS error 12 looks like a zip or apple developer sign issue, was happening since 94 also btw

danoli3:
Exit code 12 is “zip has nothing to do” according to the link above.

danoli3:
PG 0.95.0

fixing scripts for macOS deployment
fixing builds for Mac-arm64 seperate to universal gui
adding builds for linux-arm64 / linux-armv7l - gui

PR #631, opened 2025-03-28
Framework to copy phase - fix ofxSyphon issue

dimitre:
Issue described here: https://github.com/openframeworks/openFrameworks/issues/7588#issuecomment-2759580725

Is version after 99 is 1.0.0 ?


ofTheo:
Thanks @dimitre ! 
You can have 0.100.0 etc in the versioning system we are using. :)

PR #634, opened 2025-04-02
v103 - android project fixes

danoli3:
v103
# Android
- fixes copying over files of existing ofApp.h / ofApp.cpp 

# Xcode
- disable attribute setting for bin/


danoli3:
Shall fix that up in next version

